Issue #2, opened 2023-09-24
ReferenceError: React is not defined on Marker and InfoWindow components


leighhalliday:
Hello! I'm trying to get a minimal example up and running in Next.js and I'm unable to get the `Marker` or `InfoWindow` components rendering (`AdvancedMarker` works fine).

I believe the issue is due to their use of a fragment `<></>`. The compiled version of this (in `dist/index.modern.mjs` imports React as ` React$1`, and most of the references are to this, but for some reason on a fragment, it is referencing `React` without the `$1`.

```js
import React$1, { useState, useMemo, useEffect, useCallback, useContext, useLayoutEffect, forwardRef, useImperativeHandle, Children, useRef } from 'react';

const Marker = forwardRef((props, ref) => {
  const marker = useMarker(props);
  useImperativeHandle(ref, () => marker, [marker]);
  return /*#__PURE__*/React.createElement(Fragment, null);
});
function useMarkerRef() {
  const [marker, setMarker] = useState(null);
  const refCallback = useCallback(m => {
    setMarker(m);
  }, []);
  return [refCallback, marker];
}
```

The error is being displayed as:

![Screenshot 2023-09-24 at 3 34 56 PM](https://github.com/visgl/react-google-maps/assets/603921/5d9badb7-1f18-4207-b579-97fe0e1c2643)

The example app where I am running into the error can be found at https://github.com/leighhalliday/next-react-google-maps-bug

If I modify the source to refer to `<Fragment></Fragment>` (vs `<></>`), the generated dist file works correctly. This makes me think it has something to do with microbundle:

```js
{
  "build:microbundle": "microbundle -o dist/index.js -f modern,umd --globals react=React,react-dom=ReactDOM --jsx React.createElement --no-compress --tsconfig tsconfig.build.json",
}
```

PR #3, opened 2023-09-24
Specify what to use for jsxFragment option on build

leighhalliday:
This is hoping to solve the issue I encountered in https://github.com/visgl/react-google-maps/issues/2

The previous build would produce:

```js
import React$1, { useState, useMemo, useEffect, useCallback, useContext, useLayoutEffect, forwardRef, useImperativeHandle, Children, useRef } from 'react';
```

After this change it produces:

```js
import React, { useState, useMemo, useEffect, useCallback, useContext, useLayoutEffect, forwardRef, useImperativeHandle, Children, useRef } from 'react';
```

This is important because the code a fragment produces is:

```js
return /*#__PURE__*/React.createElement(React.Fragment, null);
```

Which would cause an error that said `React is not defined` when imported as `React$1`. Perhaps there is another way to fix this, but I am not super familiar with microbundle. 

Here is a microbundle issue referencing the same React reference import error: https://github.com/developit/microbundle/issues/809

This also pointed me towards this direction: https://www.fabiofranchino.com/log/how-to-use-microbundle-with-react-and-avoid-reference-error/


usefulthink:
@leighhalliday Thanks a lot for spotting this! I haven't tried it in a next.js context yet.. 
If you spot more problems with it (especially regarding SSR and such), please let us know!

usefulthink:
I just published your fix in version 0.0.4, can you retest if the problem persists?

Issue #18, opened 2023-10-12
RFC: removing hooks from the library for the 1.0 release


usefulthink:
In recent discussions we came to the conclusion that it would probably make sense to remove some of the hooks from the library for the first major release in order to limit the API surface and consolidate what we have before possibly reintroducing them if needed.

The hooks that are definitely going to stay:

 - `useMap()`
 - `useApiIsLoaded()` / `useApiLoadingStatus()`
 - `useMapsLibrary()` (with the pending change to return the library as returned by `google.maps.importLibrary`)

Trivial hooks that aren't really needed and will be removed (all of their use-cases [can be replaced with a combination of `useMapsLibrary()` with `useState()` / `useEffect()`](https://github.com/visgl/react-google-maps/blob/main/README.md#using-other-libraries-of-the-google-maps-javascript-api)):

 - `useAutocompleteService()`
 - `useDistanceMatrixService()`
 - `useElevationService()`
 - `useGeocodingService()`
 - `useMaxZoomService()`
 - `usePlacesService()` (the only thing that differs from the other hooks here is that the places-service needs an element to render attribution information to, but that is something that is going to change in future versions of the google maps API).

Finally there are the more complex hooks. Considering those, I started to think that the approach to provide all the functionality of the Maps JavaScript API _as is_ via react-hooks is not the best choice here and we should focus on providing useful abstractions solving specific common use-cases  instead.

 - `useAutocomplete()`: 
    providing a good autocomplete solution is a lot more complicated than can properly be implemented in a simple hook. As such, there would probably only be a very limited number of users this implementation would actually be useful for. The current implementation of `google.maps.places.Autocomplete` is also problematic in a react-context due to it creating DOM Elements outside of the scope of the map or the autocomplete input-element. Given the new focus on Web Components in the Maps JavaScript API, it is also likely that google will provide a WC based implementation of the autocomplete input in the future that could possibly solve most of these issues and be used instead.
 
 - `useDirectionsService()`: this seems to fit well with the Idea to solve a common and specific use-case (rendering directions-results on the map). This should probably be refactored a bit to support a wider range of use-cases.

 - `useStreetViewPanorama()`: the StreetView functionality is a unique feature of the Google Maps API and definitely shouldn't be missing from this library. We should think about refactoring this to fit the common use-cases and possibly provide an alternative component-version to just render StreetView imagery.

usefulthink:
CC @wangela, @BiniCodes, @plumdumpling

[...]

PR #20, opened 2023-10-23
refactor(hooks): remove implementation and docs for trivial hooks

mrMetalWood:
As suggested in #18 this removes six of the most trivial hooks because those don't add much value and can easily be recreated and adjusted to every users specific needs when created from scratch.


usefulthink (approved):
Thanks!

PR #21, opened 2023-10-23
docs: adjust branch naming in links, master -> main

mrMetalWood:
As we don't have a `master` branch, adjust links to use the correct name (although Github redirects them correctly)


usefulthink (approved):
Thanks for spotting this!

PR #48, opened 2023-11-02
feat: standalone examples (CodeSandbox)

mrMetalWood:
- restructure examples to be able to run the standalone on a service like CodeSandbox so that users can play with the examples right away


website/src/examples/basic-map.mdx
```diff
@@ -1,5 +1,7 @@
 # Basic Map
 
+import BrowserOnly from '@docusaurus/BrowserOnly';
+
 import App from 'website-examples/basic-map/src/app';
 
-<App />
+<BrowserOnly>{() => <App />}</BrowserOnly>
```
usefulthink, line 7: Is this the fix to get the build process to work again? I remember that at some point I thought I needed the `BrowserOnly` wrapper, but it turned out the reason for the problem was something completely different. Let's have a look if we find another way to solve this.

usefulthink:
About the styles for standalone examples, how about we just load a special stylesheet for the examples (or maybe even the full stylesheet) from something like `https://visgl.github.io/react-google-maps/styles/examples.css`. Then we wouldn't need the duplicated styles for all examples and they would still work standalone.

We'd only have to figure out how to build/deploy the file during the website-build and we would be fine. I would maybe even consider doing the same for other shared code like the `getApiKey()` function. We could just have that as a plain JS module on our site and load it from there.

usefulthink (approved):

Issue #49, opened 2023-11-03
onClick event doesn't work on Map component


alex-roc:
When I tried to get the coordinates with a onClick event in the map it doesn't show nothing. How to obtain the coordinates?

For example, this doesn't work:
```jsx
 <Map
        onClick={(event) => { console.log(event) }}
        style={{ width: '400px', height: '400px' }}
        zoom={5}
        center={{
            lat: -17.146,
            lng: -65.841
        }}
    >
        <Marker position={{ lat: values.ubicacion.lat, lng: values.ubicacion.lng }} />
</Map>
```

usefulthink:
Yeah, that is something I still have to add to the map-component. We'll add support for all map-events soon...

jabranr:
Hi, I started to look into this so I have created a PR that addresses this missing feature. https://github.com/visgl/react-google-maps/pull/52

testboui:
I liked this library very much and I would like to use it. But this issue is preventing me from doing so. Can you please provide a temporary workaround for implementing onClick. Thanks.

ByteCrak07:
This is what I did as of right now to implement onClick
`
type Position = {
  lat: number;
  lng: number;
}

interface MapsHookComponentProps {
  updatePosition: ({ lat, lng }: Position) => void
}

const MapsHookComponent = ({ updatePosition }: MapsHookComponentProps) => {
  const map = useMap();
  const coreLib = useMapsLibrary('core');

  useEffect(() => {
    if (!map || !coreLib) return;

    coreLib.event.addListener(map, 'click', function (e: any) {
      const latLng = e.latLng;
      const latitude = latLng.lat() as number
      const longitude = latLng.lng() as number

      updatePosition({ lat: latitude, lng: longitude });
    })

    return () => {
      coreLib.event.clearListeners(map, 'click')
    }
  }, [map, coreLib, updatePosition]);

  return <></>;
};

const GoogleMap = () => {
  const [position, setPosition] = useState({ lat: 0, lng: 0 })

  const updatePosition = ({ lat, lng }: Position) => {
    setPosition({ lat, lng });
  }

  return (
    <div style={{ height: '350px', width: '100%' }}>
      <APIProvider apiKey={MAPS_API_KEY}>
        <Map
          center={position}
          zoom={10}
        >
          <Marker position={position} />
          <MapsHookComponent updatePosition={updatePosition} />
        </Map>
      </APIProvider>
    </div>
  );
}
`

[...]

PR #52, opened 2023-11-04
feat: add onClick event for map

jabranr:
Hi,

This PR adds `onClick` event support for the map. This addresses the #49.

As part of this PR I have:
- [x] add an example
- [x] updated API reference
- [ ] wrote unit test coverage

**Unit tests**
I have tried to put a basic unit test but it just won't work so I would appreciate the help/feedback with this. Here is the example I tried with:

``` js
test('onClick handler is called when map is clicked', () => {
  createMapSpy.mockReset();
  const onClick = jest.fn();
  rerender(
    <GoogleMap id={'mymap'} center={center} zoom={14} onClick={onClick} />
  );

  screen.getByTestId('map').click();
  expect(onClick).toHaveBeenCalledTimes(1);
});
```

### DIFF
--- a/examples/map-on-click/README.md
+++ b/examples/map-on-click/README.md
@@ -1,0 +1,32 @@
+# Google Maps with onClick event Example
+
+![image](https://user-images.githubusercontent.com/2131246/280457336-82123c40-4779-4345-99c3-727d4f817b96.png)
+
+This is an example to show how to setup an onClick event on Google Maps Map with the `<Map/>` component of the Google Maps React
+library.
+
+## Instructions
+
+Go into the example-directory and run
+
+```shell
+npm install
+```
+
+Then start the example with
+
+```shell
+npm start
+```
+
+Running the examples locally requires a valid API key for the Google Maps Platform.
+See [the official documentation][get-api-key] on how to create and configure your own key.
+
+The API key has to be provided via an environment variable `GOOGLE_MAPS_API_KEY`. This can be done by creating a
+file named `.env` in the example directory with the following content:
+
+```shell title=".env"
+GOOGLE_MAPS_API_KEY="<YOUR API KEY HERE>"
+```
+
+[get-api-key]: https://developers.google.com/maps/documentation/javascript/get-api-key
--- a/src/components/map.tsx
+++ b/src/components/map.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable complexity */
 import React, {
   CSSProperties,
-  PropsWithChildren,
+  ComponentPropsWithRef,
   useCallback,
   useContext,
   useEffect,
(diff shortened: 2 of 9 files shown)


src/components/map.tsx
```diff
@@ -160,6 +165,14 @@ function useMapInstanceHandlerEffects(
         });
       }
 
+      if (onClick) {
+        google.maps.event.addListenerOnce(newMap, 'idle', () => {
+          google.maps.event.addListener(newMap, 'click', () => {
```
usefulthink, line 170: This is missing some code to remove the event-listener on unmount. For this, we could store the returned [`google.maps.MapsEventListener`](https://developers.google.com/maps/documentation/javascript/reference/event#MapsEventListener) object and call it's `remove()` method.

Since there's going to be a lot of events, I think this should be extracted into a separate internal hook (for example `useMapEventHandlers()` that takes care of adding and removing event-handlers based on the props. Something similar has been done in [src/components/advanced-marker.tsx](https://github.com/visgl/react-google-maps/blob/main/src/components/advanced-marker.tsx).

examples/map-on-click/README.md
usefulthink, line 1: There's going to be a lot more events added and I would like to have a single example to demonstrate them. With that in mind, could you rename the example to `map-events` or similar?

usefulthink (commented):
First of all, thanks so much for taking the time and contributing to this project. Also congratulations, you are the very first outside contributor here 🎉 

Adding the map-events was actually something that I had on my todo list for the next week, but I would be more than happy to hand this task over to you or collaborate in this PR to get it done.

I left a number of comments for you to consider with the overall vision for the event-support in mind. If you need any help, feel free to let me know, otherwise I'd try to pick it up from here and add to this PR.

usefulthink:
As for the unit-tests, that's a bit complicated since we don't actually have a real map-instance to work with (everything google-maps related comes from the [`@googlemaps/jest-mocks`](https://github.com/googlemaps/js-jest-mocks) library). 

What we need to do here is to mock emitting the events, which can get a bit complicated. 

I'd try the following:
 - In the test-environment, `google.maps.event.addListener` is actually a `jest.fn()` spy.
 - After the map is initialized, we can find the callbacks passed in by the Map component using something like
   ```tsx
   const addListenerMock = google.maps.event.addListener as jest.Mock;
   const [name, callback] = addListenerMock.calls.find(args => args[0] === 'click');`
   ```
 - after that, the `callback` can be called and verified that the function passed to the `onClick` prop is actually called

 Something similar can be done to validate that event-listeners are removed on unmount and re-render.

jabranr:
@usefulthink Thank you Martin for the feedback. I will pick it up and will revert.

usefulthink:
I'm very sorry for this, but I had to move on with this topic since it's one of the biggest bits of functionality currently missing.

The implementation for full event-support is almost ready in this PR: https://github.com/visgl/react-google-maps/pull/64/commits/5f3fe98e2faf3e5ab8be27dd87baf2071d07b5fc

I still have to add some documentation for this and add an example. Do you want to pitch in with that?

usefulthink:
Still missing tests and documentation, but the main functionality for events is now merged and available in v0.3. 
Again, apologies for rushing this but I had to get this done asap.

jabranr:
No worries! I have been quite busy recently so it was hard to take time out. Appreciate the update as you went along. I will be glad to contribute to any other parts of this repo!

PR #71, opened 2023-11-09
update ControlPosition

### LINKED ISSUE #67: ControlPosition not positioning where it should
Hello!

I was testing the library with [this simple component](https://github.com/MauricioRobayo/google-maps-test/blob/c430af0bb14012267c7832882f103e64e00fef0b/src/app/GoogleMapZoomControl.tsx):

```tsx
"use client";

import { APIProvider, ControlPosition, Map } from "@vis.gl/react-google-maps";

export default function GoogleMap() {
  const position = { lat: 53.54, lng: 10 };

  return (
    <APIProvider apiKey={process.env.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY ?? ""}>
      <div className="aspect-video max-w-lg">
        <Map
          zoom={9}
          center={position}
          disableDefaultUI
          zoomControl
          zoomControlOptions={{
            position: ControlPosition.LEFT_TOP, // <----
          }}
        />
      </div>
    </APIProvider>
  );
}
```

but the zoom control is not being position at the LEFT_TOP:

image

If I use the same constant from `google.maps.ControlPosition` it works as expected. Although they both seem to point to `17`, at runtime `google.maps.ControlPosition` seems to evaluate to `5`:

```tsx
"use client";

import { APIProvider, ControlPosition, Map } from "@vis.gl/react-google-maps";
import { useEffect, useState } from "react";

export default function GoogleMap() {
  const position = { lat: 53.54, lng: 10 };
  const [zoomControlPosition, setZoomControlPosition] = useState(
    ControlPosition.LEFT_TOP
  );

  useEffect(() => {
    setTimeout(() => {
      console.log("ControlPosition.LEFT_TOP:", ControlPosition.LEFT_TOP);
      console.log(
        "google.maps.ControlPosition.LEFT_TOP:",
        google.maps.ControlPosition.LEFT_TOP // <---- this logs 5 instead of 17
      );
      setZoomControlPosition(google.maps.ControlPosition.LEFT_TOP);
    }, 2000);
  }, []);

  return (
    <APIProvider apiKey={process.env.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY ?? ""}>
      <div className="aspect-video max-w-lg">
        <Map
          zoom={9}
          center={position}
          disableDefaultUI
          zoomControl
          zoomControlOptions={{
            position: zoomControlPosition,
          }}
        />
      </div>
    </APIProvider>
  );
}
```

image

MauricioRobayo:
Updates the ControlPosition enum according to this: https://github.com/visgl/react-google-maps/issues/67#issuecomment-1803588928

Fixes #67

### DIFF
--- a/src/components/map-control.tsx
+++ b/src/components/map-control.tsx
@@ -8,6 +8,10 @@ type MapControlProps = PropsWithChildren<{
   position: ControlPosition;
 }>;
 
+/**
+ * Copy of the `google.maps.ControlPosition` constants.
+ * They have to be duplicated here since we can't wait for the maps API to load to be able to use them.
+ */
 export enum ControlPosition {
   TOP_LEFT = 1,
   TOP_CENTER = 2,
--- a/src/components/map-control.tsx
+++ b/src/components/map-control.tsx
@@ -12,37 +12,39 @@ type MapControlProps = PropsWithChildren<{
  * Copy of the `google.maps.ControlPosition` constants.
  * They have to be duplicated here since we can't wait for the maps API to load to be able to use them.
  */
-export enum ControlPosition {
-  TOP_LEFT = 1,
-  TOP_CENTER = 2,
-  TOP = 2,
-  TOP_RIGHT = 3,
-  LEFT_CENTER = 4,
-  LEFT_TOP = 5,
-  LEFT = 5,
-  LEFT_BOTTOM = 6,
-  RIGHT_TOP = 7,
-  RIGHT = 7,
-  RIGHT_CENTER = 8,
-  RIGHT_BOTTOM = 9,
-  BOTTOM_LEFT = 10,
-  BOTTOM_CENTER = 11,
-  BOTTOM = 11,
-  BOTTOM_RIGHT = 12,
-  CENTER = 13,
-  BLOCK_START_INLINE_START = 14,
-  BLOCK_START_INLINE_CENTER = 15,
-  BLOCK_START_INLINE_END = 16,
-  INLINE_START_BLOCK_CENTER = 17,
-  INLINE_START_BLOCK_START = 18,
-  INLINE_START_BLOCK_END = 19,
-  INLINE_END_BLOCK_START = 20,
-  INLINE_END_BLOCK_CENTER = 21,
-  INLINE_END_BLOCK_END = 22,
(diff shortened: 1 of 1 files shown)


src/components/map-control.tsx
```diff
@@ -8,35 +8,36 @@ type MapControlProps = PropsWithChildren<{
   position: ControlPosition;
 }>;
 
-/**
- * Copy of the `google.maps.ControlPosition` constants.
- * They have to be duplicated here since we can't wait for the maps API to load to be able to use them.
- */
 export enum ControlPosition {
```
usefulthink, line 11:
```suggestion
/**
 * Copy of the `google.maps.ControlPosition` constants.
 * They have to be duplicated here since we can't wait for the maps API to load to be able to use them.
 */
export enum ControlPosition {
```

usefulthink:
Awesome, thanks a lot. Merging this right away.

usefulthink:
Dang, I just noticed that the values are duplicated for aliases (`TOP` === `TOP_CENTER`), so this doesn't work as an enum. We could either remove the duplicates and keep the enum or declare this as an const object. What do you think?

MauricioRobayo:
I don't love enums, so I'm fine with making it an object if that works for you. Will come back to this later today.

usefulthink:
Perfect, let's do this.

src/components/map-control.tsx
```diff
@@ -12,32 +12,39 @@ type MapControlProps = PropsWithChildren<{
...
+  INLINE_START_BLOCK_START: 18,
+  INLINE_START_BLOCK_END: 19,
+  INLINE_END_BLOCK_START: 20,
+  INLINE_END_BLOCK_CENTER: 21,
+  INLINE_END_BLOCK_END: 22,
+  BLOCK_END_INLINE_START: 23,
+  BLOCK_END_INLINE_CENTER: 24,
+  BLOCK_END_INLINE_END: 25
+} as const;
+export type ControlPosition =
```
MauricioRobayo, line 46: Not sure about the naming convention here, should I use a different name to distinguish between the type and the object?

usefulthink: I've seen that recommended elsewhere and it seems to work exactly like the enums, so let's do the same name..

PR #82, opened 2023-11-10
Directions example

leighhalliday:
Here's an example which uses the `useMapsLibrary` hook to load the `routes` library. Then we can combine `DirectionsService` with `DirectionsRenderer` to show the selected route on the map, and allow the user to update which route is being rendered.

Feel free to make/suggest any changes that you'd like made.


usefulthink (approved):

usefulthink:
It's deployed now, but the API key won't allow using the routes API right now, I'm on it. #85

junwen-k:
Hi, I was wondering why not use the `useDirectionsService` hook instead? I can't seem to find any code examples for that hook.

usefulthink:
@junwen-k Mostly because we were still thinking about removing those hooks for the first release to keep the surface area of the library smaller. It could still be added later after some "user-space" experimentation.

See also here: https://github.com/visgl/react-google-maps/issues/18

junwen-k:
@usefulthink Thank you for the quick reply! I was just slightly confused because `useDirectionsService` hook is written in the [documentation](https://visgl.github.io/react-google-maps/docs/api-reference/hooks/use-directions-service) and I thought that was supposed to be the official recommended way. Now I know that not using it was intentional, thanks :)

PR #86, opened 2023-11-12
Adding example for marker clustering

leighhalliday:
Here's an example for marker clustering. Because it uses `AdvancedMarker`, I had to add an env var for the `mapId` as well, but I've updated the README to mention that.

It should note that this uses an additional package `@googlemaps/markerclusterer` which means I couldn't add it as an example onto the website.

Happy to make any changes you might see fit!

ps - I also fixed a link in the `change-map-styles` example.


usefulthink:
Awesome, thank you so much! I'll have a look tomorrow. 

Regarding the mapId, even though they look funny, they are not to be considered secrets and can be shared between projects without problem (except for the problem to find out where the mapId was defined...) 
For other examples I just copied some mapIds we had created for other projects and didn't bother to put them into env-variables, I would suggest you do the same here. If you want, you can copy the mapId from the `markers-and-infowindows` example.

leighhalliday:
Awesome, great idea @usefulthink I've updated the example to use the mapId from the other example and have removed references to that ENV var.

usefulthink (approved):

geoffreygarrett:
Hi, I'm encountering a bit of a frustrating issue regarding getting clustering working efficiently with an `InfoWindow`. 

No matter how I attempt to integrate an InfoWindow, that triggers on state change of the current active item, ALL my markers rerender. This only happens when the clustering is used.

Here's a minimal example, where I add a window to your existing example:

https://codesandbox.io/p/sandbox/mutable-frost-hfvnqw

You'll notice that even with attempted memoization, all my markers rerender.

Specifically, the strategy I used was:

```tsx
    // ...
    const [activeItem, setActiveItem] = useState<Point>(null);
    
    // ...
     const handleItemClick = useCallback((item: Point) => {
        setActiveItem(item);
    }, []);

    // ...
        return (
        <>

            {points.map(point => (
                <AdvancedMarker
                    position={point}
                    onClick={() => handleItemClick(point)}
                    key={point.key}
                    ref={marker => setMarkerRef(marker, point.key)}>
                    <span className="tree">🌳</span>
                </AdvancedMarker>
            ))}
            {activeItem &&
                (
                    <InfoWindow
                        position={activeItem}
                        pixelOffset={{width: 0, height: -30}}>
                        {activeItem.name}
                    </InfoWindow>

                )

            }

        </>
    );
```

I'm not the most experienced with react, so perhaps it's a misunderstanding on my behalf. I've tried many alternatives over the past 2 days, and I'm almost at the conclusion that this isn't possible to avoid rerenders when using the click handler of a marker when it belongs to a cluster.

PR #108, opened 2023-11-29
docs: add marker clustering example to website

wangela:
Add the Marker Clustering example from #86 to the website


usefulthink:
Unfortunately, we can't add examples that are using their own dependencies (like `@googlemaps/markerclusterer` in this case) to the website right now, since the dependencies from the `examples/*` subdirectories aren't automatically installed and are missing during the build.

But I take this as a note that we should solve this or at least make sure those examples are listed on the website (even if you'll need to open them in codesandbox to see them).

wangela:
> Unfortunately, we can't add examples that are using their own dependencies (like `@googlemaps/markerclusterer` in this case) to the website right now, since the dependencies from the `examples/*` subdirectories aren't automatically installed and are missing during the build.
> 
> But I take this as a note that we should solve this or at least make sure those examples are listed on the website (even if you'll need to open them in codesandbox to see them).

oh woops, just saw this after I installed the dependency to the website folder! I'll leave it up to you whether this is acceptable or if we should close the PR without merging.

wangela:
After offline discussion, decided to link to CodeSandbox for examples with external dependencies. Updated /examples README with this guidance and applied to marker-clustering example.

wangela:
Also confirmed that adding an external link automatically adds the "external" styling to the link. 
screenshot

usefulthink:
That is so cool, thanks! I allowed myself to pull that regex masterpiece into it's own function.

PR #119, opened 2023-12-01
docs: minor changes

usefulthink:


wangela (changes requested):
Thanks for these fixes! Would you also update the website index page Javascript -> JavaScript in L59? Since it's large text it's quite visible. https://github.com/visgl/react-google-maps/blob/usefulthink-patch-1/website/src/pages/index.jsx#L59

PR #130, opened 2023-12-13
docs: correct links to the hooks source code files

WanderingBrooks:
Tried to find the source code for the hooks but the links point to tsx files but the hooks are ts files.

Excited to see where this project goes :D!


usefulthink:
Awesome, thanks!

PR #135, opened 2023-12-20
fix: avoid re-render on every importLibrary() call

thekip:
`importLibrary()` depends on `loadedLibraries` but that is not stated in the deps array. This cause check: 

```js
if (loadedLibraries[name]) {
  return loadedLibraries[name];
}
```

Not working, because `loadedLibraries` always empty and therefore every `importLibrary()` call even for libraries which are already loaded cause full re-render of all components depending on map context.

### DIFF
--- a/src/components/api-provider.tsx
+++ b/src/components/api-provider.tsx
@@ -142,7 +142,7 @@ function useGoogleMapsApiLoader(props: APIProviderProps) {
 
       return res;
     },
-    []
+    [loadedLibraries]
   );
 
   useEffect(


thekip:
@usefulthink ping

wangela:
@thekip Please expect delays in maintainer response times during the holiday season.

thekip:
All good, just wasn't sure do maintainers subscribed for all new opened pull reqests.

usefulthink:
I am :) 

And I'm usually very actively watching this repository. Just happened to recover from intercontinental travel over the past days..

Anyway. Thanks for the pr! That looks pretty clearly like an oversight on my end. Will merge and release it right away.

Happy holidays 🎄🎁🌟

Issue #137, opened 2023-12-21
[Bug] Invalid/Expired API key crashes the page


iqbalio:
### Description

If the API key is expired/wrong. Then the page crashes once we move away from the map page. It throws the following error

`Cannot read properties of undefined (reading 'remove')`

### Steps to Reproduce

https://codesandbox.io/p/sandbox/gracious-rain-slhtvx

1. Open the above sandbox and click on "Map" link
2. Map page will open and It will show "Oops! Something went wrong." as the API key is wrong.
3. Click on Home link to go back to home page
4. Click on Map again

Now after this If you move again to home page or any other route, It'll crash the page.

### Environment

- Library version: 0.4.1
- Google maps version: weekly
- Browser and Version: Chrome 120
- OS: Windows 11

usefulthink:
This seems to be coming from the effect-cleanup for the `useMapsEvents` hook. 
Needs some further investigation. 

We should also implement the `window.gm_authFailure` callback in the API-loader so we can properly react to problems with the API-key.

usefulthink:
Interesting:

https://github.com/visgl/react-google-maps/blob/69bade05004234c0777d92a3be541369ec438e7a/src/components/map/use-map-events.ts#L62-L77

For the error to happen, `listener` has to be `undefined`, which - at least according to the google maps docs can't happen. I verified that in the debugger and indeed there seems to be a situation where the `google.maps.Map` instance has an empty function instead of `addListener` which isn't documented anywhere. i'll try to replace that with `google.maps.events.addListener` to see if that is more reliable.

usefulthink:
With #158 and #165 this should be taken care of. Will release a new Version later today.

PR #158, opened 2024-01-17
fix: don't use potentially unreliable addListener functions

usefulthink:
As discovered in #137, there are apparently situations where the `.addListener()` function that components inherit from the MVCObject class will return undefined in certain edge cases, like when the Map is rendering the fallback-screen for an invalid API-key.

Therefore we will use the hopefully more stable `google.maps.event.addListener` method everywhere.

### DIFF
--- a/src/libraries/google-maps-api-loader.ts
+++ b/src/libraries/google-maps-api-loader.ts
@@ -27,7 +27,7 @@ const MAPS_API_BASE_URL = 'https://maps.googleapis.com/maps/api/js';
 export class GoogleMapsApiLoader {
   /**
    * Loads the Google Maps API with the specified parameters.
-   * Since the maps library can only be loaded once per page, this will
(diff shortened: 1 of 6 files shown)


src/libraries/google-maps-api-loader.ts
```diff
@@ -39,6 +39,10 @@ export class GoogleMapsApiLoader {
     const libraries = params.libraries ? params.libraries.split(',') : [];
     const serializedParams = this.serializeParams(params);
 
+    // note: if google.maps.importLibrary was defined externally, the params
+    //   will be ignored. If it was defined by a previous call to this
+    //   method, we will check that the key and other paramters have not been
```
mrMetalWood, line 44:
```suggestion
    //   method, we will check that the key and other parameters have not been
```

mrMetalWood (approved):

PR #180, opened 2024-01-26
Update get-started.md

TrustDec:


usefulthink:
fair point :D We could also add that directly to the readme while we're at it.

Issue #175, opened 2024-01-23
How to make clusterization map + infoWindows for markers ?[Feat]


DomnitskiyOleg:
### Target Use Case

How to make clusterization map + infoWindows for markers ?

### Proposal

-------

anthfgreco:
I tried implementing this as well but to no avail. The map freezes for a few seconds before the info window pops up. If anybody has a working example it'd be greatly appreciated.

I tried working with this example that doesn't have info windows: [Code SandBox Marker Clustering](https://codesandbox.io/p/devbox/github/visgl/react-google-maps/tree/main/examples/marker-clustering)

mfmanca2000:
I have a solution for adding both clustering and infowindows but it's quite awful and I still have a bug: when clicking on an Advanced Marker, two InfoWindows open up. One it's the correct one, another is an empty one that never goes away, it's just hidden when you click its closing button.

Online some others had the same issue with other libraries as well. Their solutions didn't work for me: some removed the React.StrictMode wrapper in index.js and some others switched to using InfoWindowF instead of InfoWindow.

Anyway my solution was to have a state with an array that will contain the id of the markers that should have a visible infowindow.
It's based on the clustering example (https://codesandbox.io/s/github/visgl/react-google-maps/tree/main/examples/marker-clustering) plus basically...
1.  the array in the state
2. the onClick on the marker that will add or remove a key in this array
3. the InfoWindow that will be drawn only if the array contains the key
4. The anchor of the InfoWindow that should get the marker from the markers array (i.e. "markers" in the clustering example)
5. the onCloseClick handler on the infowindow

```
const [openFlagsArray, setOpenFlagsArray] = useState([]);
...
return (
        <>
            {locations.map((loc) => (
                <div key={'div_' + loc.$id}>

                    <AdvancedMarker
                        position={{ lat: loc.latitude, lng: loc.longitude }}
                        key={loc.$id}
                        title={loc.name}
                        ref={(marker) => setMarkerRef(marker, 'm_' + loc.$id)}
                        onClick={() => {                                                                            
                            let i = openFlagsArray ? openFlagsArray.indexOf(loc.$id) : -1;
                            if (i != -1) {
                                setOpenFlagsArray((prev) => {                                    
                                    return prev.filter(flag => {
                                        return flag != loc.$id
                                    })                                                                        
                                })
                            } else {
                                setOpenFlagsArray((prev) => {                                    
                                    return [...prev, loc.$id];                                                                        
                                })
[...]

Issue #109, opened 2023-11-29
[Bug] InfoWindow renders twice after content load


Taltalonix:
### Description

When using the InfoWindow component it renders twice, once empty and then when the children components are rendered it refreshes.  

https://github.com/visgl/react-google-maps/assets/35221237/4b31b633-37c1-4fa5-aa18-fa635f9f1ab8

Wasn't sure if it's relevant enough to create a pull request, but this fixed the problem:
```tsx
/**
 * Props for the Info Window Component
 */
export type InfoWindowProps = google.maps.InfoWindowOptions & {
    onCloseClick?: () => void;
    anchor?: google.maps.Marker | google.maps.marker.AdvancedMarkerElement | null;
};

/**
 * Component to render a Google Maps Info Window
 */
export const InfoWindow = (props: PropsWithChildren<IMyInfoWindowProps>) => {
    const { children, anchor, onCloseClick, ...infoWindowOptions } = props;
    const map = useContext(GoogleMapsContext)?.map;

    const infoWindowRef = useRef<google.maps.InfoWindow | null>();

    const [contentContainer, setContentContainer] =
        useState<HTMLDivElement | null>(null);

    // create infowindow once map is available
    useEffect(() => {
        if (!map) return;

        if (!infoWindowRef.current) {
            infoWindowRef.current = new google.maps.InfoWindow(infoWindowOptions);
        }

        // Add content to info window
        const el = document.createElement('div');
        infoWindowRef.current.setContent(el);

        if (onCloseClick) {
            infoWindowRef.current.addListener('closeclick', () => {
                onCloseClick();
            });
        }

        setContentContainer(el);

        // Cleanup info window and event listeners on unmount
        return () => {
            if (infoWindowRef.current) {
                google.maps.event.clearInstanceListeners(infoWindowRef.current);
                infoWindowRef.current.close();
            }
            infoWindowRef.current = null;
            el.remove();

            setContentContainer(null);
        };
    }, [map, children, anchor]);

    // Open info window after content container is set
    useEffect(() => {
        if (contentContainer && infoWindowRef.current && anchor) {
            infoWindowRef.current.open({ map, anchor });
        }
    }, [contentContainer, infoWindowRef, anchor, map]);

    return (
        <>{contentContainer !== null && createPortal(children, contentContainer)}</>
    );
};
```

https://github.com/visgl/react-google-maps/assets/35221237/c203c932-bb5f-4525-bb57-e65ffef194db

### Steps to Reproduce

[...]

PR #181, opened 2024-01-26
feat: improved state-handling implementation

usefulthink:

PR #185, opened 2024-01-30
fix(infowindow): fix double rendering and fix eslint warnings

mrMetalWood:
This should fix the double rendering issues of the infowindow causing problems (#109 #175)

- Restructure infowindow to handle options update and callback update without having to re-render the whole infowindow which would cause flickering
- Handle className update in the AdvancedMarker component without having to re-render the marker which would cause flickering of the marker
- fix a few eslint warning by deliberately disabling the `react-hooks/exhaustive-deps` on a few lines

### DIFF
--- a/src/components/info-window.tsx
+++ b/src/components/info-window.tsx
@@ -25,7 +25,7 @@ export const InfoWindow = (props: PropsWithChildren<InfoWindowProps>) => {
   const {children, anchor, onCloseClick, ...infoWindowOptions} = props;
   const map = useContext(GoogleMapsContext)?.map;
 
-  const infoWindow = useRef<google.maps.InfoWindow | null>(null);
+  const infoWindowRef = useRef<google.maps.InfoWindow | null>(null);
   const [contentContainer, setContentContainer] =
     useState<HTMLDivElement | null>(null);
 
@@ -39,7 +39,7 @@ export const InfoWindow = (props: PropsWithChildren<InfoWindowProps>) => {
     const el = document.createElement('div');
     newInfowindow.setContent(el);
 
-    infoWindow.current = newInfowindow;
+    infoWindowRef.current = newInfowindow;
     setContentContainer(el);
 
     // Cleanup info window and event listeners on unmount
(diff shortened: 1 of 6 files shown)


src/components/info-window.tsx
```diff
@@ -19,38 +25,76 @@ export const InfoWindow = (props: PropsWithChildren<InfoWindowProps>) => {
   const {children, anchor, onCloseClick, ...infoWindowOptions} = props;
   const map = useContext(GoogleMapsContext)?.map;
 
+  const infoWindow = useRef<google.maps.InfoWindow | null>(null);
```
usefulthink, line 28: Since it's not the infowindow itself here, can we rename this to `infoWindowRef`?

src/components/info-window.tsx
```diff
@@ -19,38 +25,76 @@ export const InfoWindow = (props: PropsWithChildren<InfoWindowProps>) => {
...
-  }, [map, children, anchor]);
+    // We don't want to re-render a whole new infowindow
+    // when the options change to prevent flickering.
+    // Update of infoWindow options is handled in the useEffect below.
+    // Excluding infoWindowOptions from dependency array on purpose here.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [map, children]);
+
+  // Update infoWindowOptions
+  useEffect(() => {
```
usefulthink, line 62: this may or may not cause problems: due to destructuring (line 25) `infoWindowOptions` will be considered changed for every render. In #181 we'll add a [`useDeepCompareEffect`](https://github.com/visgl/react-google-maps/pull/181/files#diff-5118501a9135d733bb840b77dcafcfaf23f121e4eae24be4d630ef1e968a1dc2) that we can use here instead.

I'd say leave it like this for now.

usefulthink (approved):
Awesome, only one to go :D

src/components/info-window.tsx
mrMetalWood: Sure :)

PR #206, opened 2024-02-08
refactor(use-map-instance): remove unused import

shuuji3:
Hi, this is a simple fix to address one ESLint warning.

ref. https://github.com/visgl/react-google-maps/actions/runs/7817624474/job/21325966198#step:5:25


usefulthink (approved):
Thanks!

PR #220, opened 2024-02-13
feat: add example for drawing tools

mrMetalWood:


examples/drawing/src/undo-redo.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  const pinnedStatesSnapshot = JSON.stringify(
+    state.now.map((overlay: Overlay) => overlay.pinnedState)
+  );
+
+  useEffect(() => {
+    if (!drawingManager) return;
+
+    const addUpdateListener = (eventName: string, drawResult: DrawResult) => {
+      google.maps.event.addListener(drawResult.overlay, eventName, () => {
```
usefulthink, line 40: is it neccessary to remove the listeners at some point?

examples/drawing/src/undo-redo.tsx
```diff
@@ -0,0 +1,147 @@
...
+  isPolygon,
+  isPolyline,
+  isRectangle
+} from './types';
+
+interface Props {
+  drawingManager: google.maps.drawing.DrawingManager | null;
+}
+
+export const UndoRedo = ({drawingManager}: Props) => {
```
usefulthink, line 21: Could this be renamed to `UndoRedoControl`? 

Another related Idea would be to extract the effects from here into reducer.ts (and rename that to undo-redo.ts), so there's one place where both data definition and manipulation happen. This file would then be only the UI parts of Undo/Redo.

examples/drawing/src/undo-redo.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  const [state, dispatch] = useReducer(reducer, {
+    past: [],
+    now: [],
+    future: []
+  });
+
+  const overlaysShouldUpdate = useRef<boolean>(false);
+
+  const pinnedStatesSnapshot = JSON.stringify(
```
usefulthink, line 32: this would benefit from a comment in why this is here and what the pinnedState is

examples/drawing/src/undo-redo.tsx
```diff
@@ -0,0 +1,147 @@
...
+export const UndoRedo = ({drawingManager}: Props) => {
+  const map = useMap();
+
+  const [state, dispatch] = useReducer(reducer, {
+    past: [],
+    now: [],
+    future: []
+  });
+
+  const overlaysShouldUpdate = useRef<boolean>(false);
```
usefulthink, line 30: rename to `overlaysShouldUpdateRef`. We did that everywhere else and it immediately explains the `.current`.

examples/drawing/src/reducer.ts
```diff
@@ -0,0 +1,99 @@
...
+  isCircle,
+  isMarker,
+  isPolygon,
+  isPolyline,
+  isRectangle
+} from './types';
+
+export default function reducer(state: State, action: Action) {
+  switch (action.type) {
+    case DrawingActionKind.UPDATE_OVERLAYS: {
```
usefulthink, line 16: this would benefit from a short explanatory comment per action.

usefulthink (commented):
This is all excellent stuff, thanks a lot! I added a couple of comments and ideas, but nothing too serious...

usefulthink (approved):

PR #241, opened 2024-02-27
Update info-window.md

dancielos:
Using <AdvancedMarker> requires that the Map has a MapId. This was actually mentioned in the documentation of your <AdvancedMarker. However, in the <InfoWindow> component example when using AdvancedMarker, MapId props was not specified. 

Hence, I added the mapId prop so users will know to add their MapId when using the example provided on InfoWindow.


usefulthink:
good point, thanks.

Issue #253, opened 2024-03-10
[Bug] openOptions arent passed on to InfoWindow


Freundschaft:
### Description

im implementing a filter where im displaying the markers based on a text input.
these markers have infowindows, and when the filter changes, the infowindow gains focus and takes it away from the text input.
in InfoWindowOpenOptions there is an option [shouldFocus](https://developers.google.com/maps/documentation/javascript/reference/info-window#InfoWindowOpenOptions.shouldFocus) which somehow is not possible to add to the infowindow component, I assume this is a quick fix, passing on the option should do the trick. (https://github.com/visgl/react-google-maps/blob/main/src/components/info-window.tsx#L96)

### Steps to Reproduce

im implementing a filter where im displaying the markers based on a text input.
these markers have infowindows, and when the filter changes, the infowindow gains focus and takes it away from the text input.
in InfoWindowOpenOptions there is an option [shouldFocus]

### Environment

- Library version: 0.7.1
- Google maps version: weekly
- Browser and Version: all
- OS: all

Freundschaft:
add option to set [shouldFocus](https://developers.google.com/maps/documentation/javascript/reference/info-window#InfoWindowOpenOptions.shouldFocus)

Freundschaft:
ive added a pull request https://github.com/visgl/react-google-maps/pull/254

PR #254, opened 2024-03-10
add shouldfocus related to #253

Freundschaft:
Add option to set [shouldFocus](https://developers.google.com/maps/documentation/javascript/reference/info-window#InfoWindowOpenOptions.shouldFocus)

### DIFF
--- a/src/components/info-window.tsx
+++ b/src/components/info-window.tsx
@@ -16,7 +16,7 @@ import {GoogleMapsContext} from './map';
 export type InfoWindowProps = google.maps.InfoWindowOptions & {
   onCloseClick?: () => void;
   anchor?: google.maps.Marker | google.maps.marker.AdvancedMarkerElement | null;
-  shouldFocus?: boolean | null;
+  shouldFocus?: boolean;
 };
 
 /**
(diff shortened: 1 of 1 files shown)


usefulthink:
Awesome, thanks! I think the `shouldFocus` prop needs to be added to the `InfoWindowProps` type, since that only includes the `google.maps.InfoWindowOptions`, not the `google.maps.InfoWindowOpenOptions`.

Also please check the formatting with prettier (`npm run test:prettier` to check and `npx prettier --write ./src/components/info-window.tsx` to fix).

Freundschaft:
whoops youre right ill fix and resubmit

usefulthink:
> Could it also be considered to add InfoWindowOpenOptions as a prop entirely in case the interface changes in the future?

I don't think that makes a lot of sense, since the open-options have to always be considered separately as you did here. So new options would just cause unexpected behaviour until they are implemented.

src/components/info-window.tsx
```diff
@@ -16,13 +16,15 @@ import {GoogleMapsContext} from './map';
 export type InfoWindowProps = google.maps.InfoWindowOptions & {
   onCloseClick?: () => void;
   anchor?: google.maps.Marker | google.maps.marker.AdvancedMarkerElement | null;
+  shouldFocus?: boolean | null;
```
usefulthink, line 19: I think we can leave away the `| null` here. Optional and boolean should be enough.

```suggestion
  shouldFocus?: boolean;
```

usefulthink (commented):
except for the type it's perfect.

PR #265, opened 2024-03-18
Default style

laotoutou:


docs/get-started.md
```diff
@@ -45,6 +45,7 @@ import {APIProvider, Map} from '@vis.gl/react-google-maps';
 const App = () => (
   <APIProvider apiKey={API_KEY}>
     <Map
+      style={{width: '500px', height: '700px'}}
```
usefulthink, line 48: Let's make it fullscreen :) Also fits better with the use of gestureHandling and disableDefaultUI

```suggestion
      style={{width: '100vw', height: '100vh'}}
```

usefulthink:
Thanks a lot!

PR #281, opened 2024-03-26
Export ApiLibraryKeys for use in client code

man-trackunit:
This way clients won't have to make their own copy ApiLibrary keys. Not sure if there's a better name than `ApiLibraryKeys`.

**Alternatively** a client could extract the type with something like:
```
type FirstArg<F extends (...args: never) => unknown> = Parameters<F>[0];
type ApiLibrary = FirstArg<typeof useMapsLibrary>;
```
But I'd consider that a _hack_. Better to just export the type imho. 🙏

### DIFF
--- a/src/hooks/use-maps-library.ts
+++ b/src/hooks/use-maps-library.ts
@@ -18,8 +18,10 @@ interface ApiLibraries {
   visualization: google.maps.VisualizationLibrary;
 }
 
+export type ApiLibraryKeys = keyof ApiLibraries;
+
 export function useMapsLibrary<
-  K extends keyof ApiLibraries,
+  K extends ApiLibraryKeys,
   V extends ApiLibraries[K]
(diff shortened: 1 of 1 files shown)


usefulthink:
This is actually a fix for a shortcoming in the `@types/google.maps` package. I already raised that issue in their tracker (googlemaps/js-types#95) but haven't heard back so far. It could be that this will get solved for `@types/google.maps` and you can import it from there. I would much prefer it that way.

If we implement this in this library, I would prefer to export the entire map as well, but also would like to find a better name for it first (maybe `GoogleMapsLibraryMap` / `GoogleMapsLibraryName`? what do you think?)

man-trackunit:
Hmm. Don't have a good suggestion for naming but your reasoning makes sense. I'll close this PR and hope you get through to `@types/google.maps` thanks!

PR #290, opened 2024-03-31
Fix map controls crashing when invalid key is provided

### LINKED ISSUE #276: [Bug] Adding <MapControl /> to a <GoogleMap /> with invalid/expired key make component crash
### Description

When GoogleMap component contains a MapControl and APIProvider key is invalid or expired, the app crash.

In https://github.com/visgl/react-google-maps/blob/750acb981f5c250fbc6fea3fda59cdf8eb00f60d/src/components/map-control.tsx#L61

indexOf is crashing if controls.getArray() is undefined.

### Steps to Reproduce

Create a GoogleMap containing a MapControl with an expired api key.

### Environment

- Library version:
- Google maps version: weekly
- Browser and Version:
- OS:

### Logs

_No response_

MauricioRobayo:
I'm having the same issue reported on #276.

I tested it and adding optional chaining before calling indexOf fixes it.

closes #276

### DIFF
--- a/src/components/map-control.tsx
+++ b/src/components/map-control.tsx
@@ -58,7 +58,7 @@ export const MapControl = ({children, position}: MapControlProps) => {
     controls.push(controlContainer);
 
     return () => {
-      const index = controls.getArray().indexOf(controlContainer);
+      const index = controls.getArray()?.indexOf(controlContainer);
       controls.removeAt(index);
     };
   }, [controlContainer, map, position]);


src/components/map-control.tsx
```diff
@@ -58,7 +58,7 @@ export const MapControl = ({children, position}: MapControlProps) => {
     controls.push(controlContainer);
 
     return () => {
-      const index = controls.getArray().indexOf(controlContainer);
+      const index = controls.getArray()?.indexOf(controlContainer);
```
usefulthink, line 61: I think I like this a bit more explicit and with a short explanation so that no one (me) will "optimize" that away again :)

```suggestion
      const controlsArray = controls.getArray();
      // controlsArray could be undefined if the map is in an undefined state (e.g. invalid API-key, see #276
      if (!controlsArray) return;

      const index = controlsArray.indexOf(controlContainer);
```

usefulthink:
Simple and works, gotta love it :)

Thanks a lot for the fix <3

PR #297, opened 2024-04-03
Add eslint rule to force devs to memoize values sent to context providers

rodrigofariow:
Relates to https://github.com/visgl/react-google-maps/pull/287

Also added a very basic, not so opinionated settings.json and extensions.json file with prettier correctly configured.
Sounds like a good thing to have on the git repo without causing issues

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -43,7 +43,7 @@ module.exports = {
 
         'no-undef': 'off',
         'react-hooks/rules-of-hooks': 'error',
-        'react-hooks/exhaustive-deps': 'warn',
+        'react-hooks/exhaustive-deps': 'error',
         'react/jsx-no-constructed-context-values': 'error',
 
         // We use function hoisting to put exports at top of file
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,3 +1,0 @@
-{
-  "recommendations": ["esbenp.prettier-vscode"]
-}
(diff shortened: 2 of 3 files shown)


.eslintrc.js
```diff
@@ -44,6 +44,7 @@ module.exports = {
         'no-undef': 'off',
         'react-hooks/rules-of-hooks': 'error',
         'react-hooks/exhaustive-deps': 'warn',
```
rodrigofariow, line 46: Side note (non blocking):
I would also set this rule to "error" since this can be critical, specially for a library like this one.

usefulthink: Good point – lets do that as well...

I wish there would be a bit more fine-grained control over this, we only have the option to disable the entire rule where we need to. Would be awesome if we could tell eslint to just ignore one specific variable used in a hook.

usefulthink (approved):
Thanks a lot! I wasn't even aware of that rule.

But please remove the vscode config files, which editor and editor-configurations to use is a personal decision we don't want to interfere with. 

Also that would likely lead to endless accidental commits of peoples vscode configurations.

rodrigofariow:
> Thanks a lot! I wasn't even aware of that rule.

Yep, it's pretty awesome. We added it at my company's codebase and it has caught a lot of perf regressions. Should be more wildly used for sure!

@usefulthink 
Removed the vscode files and set the exhaustive deps rule to error.

Cheers

PR #300, opened 2024-04-08
Add mouse enter and mouse leave events on Advance Marker's Pin

aliasad106:
This allows to listen mouse over and mouse leave events on the pin of advance marker.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@vis.gl/react-google-maps",
+  "name": "@aliasad106/react-google-maps",
   "version": "0.8.3",
   "description": "React components and hooks for Google Maps.",
   "source": "src/index.ts",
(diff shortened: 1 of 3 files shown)


usefulthink:
Thanks for the PR! 

Can you explain why the listeners should be added to the `Pin` instead of the `AdvancedMarker` component? The way I see it, the `Pin` is just one of many ways to use the `AdvancedMarker` and I think we should add support for the DOM events there instead.

aliasad106:
@usefulthink  because for some reasons adding directly to AdvancedMarker doesn't work.

usefulthink:
Ok yeah, that makes sense. I think there was something with the `pointer-events` css-property that has to be adjusted to use dom-events with the markers. I'll have a look into that.

aliasad106:
@isaacpresidium I have created a PR but they never merged it so I am using my own forked one. You can use if you want.

jogoley:
Can this be merged?

usefulthink:
Thanks to @mrMetalWood, we finally have a good solution that works reliably for all types of advanced marker: #472 

Closing this as outdated.

PR #305, opened 2024-04-10
feat: dissalow map prop for markers

### LINKED ISSUE #284: [Bug] `map` prop shouldn't be valid in prop-types extending the `google.maps.*` types.
### Description

Where present, the map option can be passed to google maps API objects to instantly add it to the specified map. This collides with our internal system to add/remove elements to/from the map and shouldn't be allowed. 

Check all PropTypes extending `google.maps.*` (Marker, AdvancedMarker, InfoWindow, ...) and add `Omit<T, "map">` to the declaration.

### Steps to Reproduce

no

### Environment

- Library version: 0.8.x
- Google maps version: weekly
- Browser and Version: any
- OS: all

### Logs

_No response_

maciej-ka:
Closes #284 

- [x] check in ts examples that vs code doesn't suggest map prop
  - [x] marker
  - [x] advanced-marker
  - [x] info-window (didn't had map prop)
  - [x] map-control (didn't had map prop)

### DIFF
--- a/src/components/advanced-marker.tsx
+++ b/src/components/advanced-marker.tsx
@@ -31,7 +31,10 @@ type AdvancedMarkerEventProps = {
 };
 
 export type AdvancedMarkerProps = PropsWithChildren<
-  Omit<google.maps.marker.AdvancedMarkerElementOptions, 'gmpDraggable'> &
+  Omit<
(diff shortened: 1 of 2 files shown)


usefulthink:
Thank you very much!

PR #307, opened 2024-04-10
feat: useApi hook

maciej-ka:
While getting familiar with source, I was curious what's behind `ctx`/`context` in couple of places and would it be more clear to call it `api`? (or perhaps `mapsApi`). 

- [x] create useApi hook for convenience
- [x] apply
  - [x] in map/index 
  - [x] api-provider test
  - [x] use-maps-library hook
  - [x] use-map hook
  - [x] use-api-loading-status hook
- [ ] unit test of new hook

### DIFF
--- a/src/components/__tests__/api-provider.test.tsx
+++ b/src/components/__tests__/api-provider.test.tsx
@@ -1,4 +1,4 @@
-import React, {useContext} from 'react';
+import React from 'react';
 import {act, render, screen} from '@testing-library/react';
 import {initialize} from '@googlemaps/jest-mocks';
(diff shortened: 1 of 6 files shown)


usefulthink:
In the context of this library, I associate "api" quite strongly with the Google Maps API, so that could be a point of confusion.

I think "context" should be sufficiently clear for React devs, although it's a bit generic as a name (also note that the various contexts are in no way exposed to the outside world, so it's not really an API). The scope in which they are used is also pretty small, and everyone confronted with the code should be well aware of what is stored in the context and why.

Issue #283, opened 2024-03-27
[Bug] missing center/zoom props don't produce an error or warning


usefulthink:
### Description

When the center/zoom properties aren't set, the map can't render which can cause all sorts of problems. We should handle this somehow, for example by logging a warning to the console.

### Steps to Reproduce

no

### Environment

- Library version: 0.8.x
- Google maps version: any
- Browser and Version: all
- OS: all

BParvatkar:
Can we have a logic where if both zoom and defaultZoom aren't defined then we have a default value of defaultZoom that gets applied to the map?

usefulthink:
@BParvatkar I would like that as well, but it's difficult to pick a value not knowing what you're up to (world-map or city-map?). 

Maybe issuing a warning and defaulting to rendering a world-map combine to be the best developer experience.

maciej-ka:
How about geolocation as a failback? (which can be denied, in which case we would need failback for failback, hmmm)

usefulthink:
No, that'd be way too intrusive.

PR #308, opened 2024-04-10
feat: error on missing zoom or center

maciej-ka:
Issue #283 

- [x] console error on undefined zoom or center info
- [x] update existing tests on missing zoom/center info
- [x] rename to MapCameraState
- [x] refactor: move checking to useMapInstance()
- [x] test case which detects error

### DIFF
--- a/src/components/map/index.tsx
+++ b/src/components/map/index.tsx
@@ -37,17 +37,21 @@ export type {
 } from './use-map-events';
 
 export type MapCameraProps = {
-  center: google.maps.LatLngLiteral;
   zoom: number;
   heading?: number;
   tilt?: number;
-};
+} & (
+  | {center: google.maps.LatLngLiteral}
+  | {defaultCenter: google.maps.LatLngLiteral}
+) &
+  ({zoom: number} | {defaultZoom: number});
 
 /**
  * Props for the Google Maps Map Component
  */
 export type MapProps = google.maps.MapOptions &
   MapEventProps &
+  MapCameraProps &
   DeckGlCompatProps & {
     /**
      * An id for the map, this is required when multiple maps are present
@@ -67,8 +71,6 @@ export type MapProps = google.maps.MapOptions &
      */
     controlled?: boolean;
 
-    defaultCenter: google.maps.LatLngLiteral;
-    defaultZoom: number;
     defaultHeading?: number;
     defaultTilt?: number;
     /**
--- a/src/components/map/index.tsx
+++ b/src/components/map/index.tsx
@@ -88,6 +88,18 @@ export const Map = (props: PropsWithChildren<MapProps>) => {
     );
   }
 
+  if (props.zoom === undefined && props.defaultZoom === undefined) {
+    throw new Error(
+      'Map zoom not set. Provide either a `zoom` or a `defaultZoom` prop.'
+    );
+  }
+
+  if (props.center === undefined && props.defaultCenter === undefined) {
+    throw new Error(
+      'Map center not set. Provide either a `center` or a `defaultCenter` prop.'
+    );
+  }
+
   const [map, mapRef, cameraStateRef] = useMapInstance(props, context);
 
   useMapCameraParams(map, cameraStateRef, props);
--- a/src/components/map/index.tsx
+++ b/src/components/map/index.tsx
@@ -36,7 +36,7 @@ export type {
   MapMouseEvent
 } from './use-map-events';
 
-export type MapCameraState = {
+export type MapCameraProps = {
   center: google.maps.LatLngLiteral;
   zoom: number;
   heading?: number;
--- a/src/components/map/index.tsx
+++ b/src/components/map/index.tsx
@@ -88,18 +88,6 @@ export const Map = (props: PropsWithChildren<MapProps>) => {
     );
   }
 
-  if (props.zoom === undefined && props.defaultZoom === undefined) {
-    throw new Error(
-      'Map zoom not set. Provide either a `zoom` or a `defaultZoom` prop.'
-    );
-  }
-
-  if (props.center === undefined && props.defaultCenter === undefined) {
-    throw new Error(
-      'Map center not set. Provide either a `center` or a `defaultCenter` prop.'
-    );
-  }
-
   const [map, mapRef, cameraStateRef] = useMapInstance(props, context);
 
   useMapCameraParams(map, cameraStateRef, props);
--- a/src/components/map/index.tsx
+++ b/src/components/map/index.tsx
@@ -36,7 +36,7 @@ export type {
   MapMouseEvent
 } from './use-map-events';
 
-export type MapCameraProps = {
+export type MapCameraState = {
   center: google.maps.LatLngLiteral;
   zoom: number;
   heading?: number;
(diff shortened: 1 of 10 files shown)


src/components/map/use-map-instance.ts
```diff
@@ -35,7 +35,9 @@ export function useMapInstance(
   const {
     id,
     defaultBounds,
+    // @ts-expect-error TS complains that it's not defined in MapProps
     defaultCenter,
```
maciej-ka, line 39: Why TS thinks defaultCenter is not defined here? ... when testing on pojo typed as MapProps that prop will be there, as expected.

src/components/map/index.tsx
```diff
@@ -37,17 +37,20 @@ export type {
...
+  | {defaultCenter: google.maps.LatLngLiteral}
+) &
+  ({zoom: number} | {defaultZoom: number});
 
 /**
  * Props for the Google Maps Map Component
  */
 export type MapProps = google.maps.MapOptions &
   MapEventProps &
+  MapCameraProps &
```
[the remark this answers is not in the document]
usefulthink: `zoom` and `center` are defined in `google.maps.MapOptions` so they haven't been repeated here.

`MapCameraProps` isn't included since it is intended to be used together with the `MapCameraChangedEvent` (it's a type you can use for the details of that event), see here for example:
https://github.com/visgl/react-google-maps/blob/9618024ac9dff76b03b0cb6758e974b0361de1f0/examples/multiple-maps/src/app.tsx#L33-L41

Maybe it should be renamed to `MapCameraState` to avoid confusion with the props?

src/components/map/use-map-instance.ts
```diff
@@ -35,7 +35,9 @@ export function useMapInstance(
   const {
     id,
     defaultBounds,
+    // @ts-expect-error TS complains that it's not defined in MapProps
```
usefulthink, line 38: please don't commit those – there's nothing wrong with tests failing in draft-PRs, and it's running risk of missing one of them at some point.

src/components/map/use-map-instance.ts
usefulthink: I recreated a simplified version of the situation  [in ts-playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAMghsGA7A5gXigbwDYIFxICuAtgEYQBOANFNqnlEWZQL4DcAUB6JFALJwwAeTDAAlgHskAZygZMHKFADGEJMEoB+BvESoO7Lj2gCwAYTjFKcAAoUJYWfIBmEiQybkKLKADIoABSKWKrqlDoIyCg+AD5YACYQznCE2MBmahoUEXrRHACUnNzgJoJ2Dk78giLiUrL+phZWFLb2jkXOhEjKtUhQzgFgDKbljvlYwcp1wAlJKWkZYd5yUGCcLEA). 

The error message _"Property 'defaultCenter' does not exist on type 'MapProps'."_, is a bit misleading, but its true: You're using union-types, and those only contain the fields that are present in all variations of the union (in case of `{center: X} | {defaultCenter: X}` that is the empty object). The center prop works since it's already defined in the `google.maps.MapOptions` type. 

What would work here is to specify the union as `{center: LatLng, defaultCenter?:LatLng} | {center?: LatLng, defaultCenter: LatLng}`, [see here](https://www.typescriptlang.org/play?ssl=20&ssc=1&pln=20&pc=13#code/C4TwDgpgBAMghsGA7A5gXigbwDYIFxICuAtgEYQBOANFNqnlEWZQL4DcAUB6JFALJwwAeTDAAlgHskAZygZMHKFABmEiQH4GTchU5KAxhCTBKm2AmQoO7Lj2gCwAYTjFKcAAoUJYWRgAUiliGxpQM8IioNAAmEMpwhNjAjkYmFJrhlixQAD5BKaZ4GZFQMXEJSfkUYRaoLBwAlJzc4PaCnt6+-IIi4lKyAGRdTi5u7T5NyoRI+r1IKn5gDA5j0vVYgfp9wHkh1CWx8YnJu1kYYJx1HMp+mMGpDDj4AIw0dCh4TywsjVc3pYcVXYPXDAD6veifb6ca6YVQSBifRpAA). The union type then resolves to `{center?:LatLng, defaultCenter?:LatLng}`.

usefulthink: [Here's an alternative](https://www.typescriptlang.org/play?ssl=26&ssc=1&pln=26&pc=13#code/C4TwDgpgBAMghsGA7A5gXigbwDYIFxICuAtgEYQBOANFNqnlEWZQL4DcAUF6JFALJwwAeTDAAlgHskAZygZMHKFABmEiQH4GTchU5KAxhCTBKm2AmQoO7buGgBhKcAoTs2CABN7cYpTgAFFzBZeUNjSgZ4RFQaDwhlOEJsYHsjEwozJAgAN1ZOHmgAVSR9Jxc3T29fCgCgkKw4hKSUtIjzaJQaMPTMnLyuDgL+QSq-QIlguSgACkUoR2Ny9y8fMbqoAB8oYtLF12XRmvHgjgBKTkG7YbBj+oFhUUkZKAAya8PaiekLjmVCEvEUhU0zADHut1OWDmu2kwCw3UosXiiWSqXCFBYUzAnBYXGU00wCIoDBw+AAjDQ6Cg8GSWCxzr8CY0US10STcMAaZT6LT6Zx8ZhVBIGLTzkA) that produces slightly more readable error-messages.

usefulthink:
Thanks again for your initiative!
One thing I would love to see added is a warning or error-message triggered in `useMapInstance()` when attempting to create a map without valid center/zoom for JS users.

src/components/map/use-map-instance.ts
maciej-ka: Seems it's a puzzle if TS can report this error as "either center or defaultCenter has to be provided", so far warnings had only one of two prop names. This and unsolved @ts-expect-error ... I will switch to console.warn solution.

usefulthink: I think we can go with the [last option](https://github.com/visgl/react-google-maps/pull/308#discussion_r1560848530), I like the Idea of being able to statically warn about missing required props, even if the error-message is a bit clunky. I'll try a few more things to see if it gets any better...

src/components/map/index.tsx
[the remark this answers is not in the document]
maciej-ka: Thanks. I would add that rename.

maciej-ka:
Thanks a lot for guidance and codepens. The console error is ready here.
In TS `Controlled`/`Uncontrolled` sounds like a way to go, but perhaps in separate PR?

usefulthink:
I changed a few things around, I hope you don't mind. Didn't want to bother you with this...

 - I reverted the change of the `MapCameraProps` to `MapCameraState`, I remembered why it was called that way to begin with, it's the type you can use if you want to store the prop-values related to the camera in a 'controlled component' type of situation. So I feel MapCameraProps is a fitting name.
 - I changed the behaviour from throwing an exception to logging a warning as discussed in #283, the map will now render something even without any props specified. I feel like this will be the best developer experience.
 - There's another case we didn't look at, and that is providing `defaultBounds` instead of `center`/`zoom` values. Added support for that as well

PR #332, opened 2024-04-30
docs: use a different example for `useMapsLibrary` in main readme

usefulthink:


benshabatnoam:
Much better 😄  not it's super clear to me!

Issue #105, opened 2023-11-28
[Feat] Implement `<Polyline>` component.


benschlegel:
### Target Use Case

Have a react component for `google.maps.Polyline` similar to `<Marker>`/the other components.

Is this something this project plans on doing? I'm asking because I was planning on making a PR that implements this, but wanted to ask if this is in the scope of this project before putting in work into the PR.

### Proposal

```jsx
<Polyline {...props />
```

JoshuaSkootsky:
I am also interested in this!

shaunmitchellve:
I would like to +1 this. I've just recently had to do this for displaying multiple routes on the map with click handlers, since the directionrenderer doesn't expose the click event.

guilhermenunes09:
This would be really helpful, specially dealing with GeoJson data. That being said, is there a workaround for using its libraries, I couldn't find it yet. Thanks!

nearwood:
> This would be really helpful, specially when dealing with GeoJson data. That being said, is there a workaround for using its libraries, I couldn't find it yet. Thanks!

What coincidence, I just started using this library today for a side project and needed the same.

Here's what I used (`<APIProvider>` is in a parent component): 

```ts
import React from 'react';
import { Map, useMap, useMapsLibrary } from '@vis.gl/react-google-maps';

export default function MyMap() {
  const map = useMap();
  const maps = useMapsLibrary("maps");
  const position = { lat: 0, lng: -180 };

  const flightPlanCoordinates = [
    { lat: 37.772, lng: -122.214 },
    { lat: 21.291, lng: -157.821 },
    { lat: -18.142, lng: 178.431 },
    { lat: -27.467, lng: 153.027 },
  ];

  if (!maps) {
    return null;
  }

  const flightPath = new maps.Polyline({
    path: flightPlanCoordinates,
    geodesic: true,
    strokeColor: "#FF0000",
    strokeOpacity: 1.0,
    strokeWeight: 2,
  });

  flightPath.setMap(map);

  return (
    <Map
      zoom={3}
      center={position}
      gestureHandling={'greedy'}
      disableDefaultUI={true}
    />
  );
}

```

guilhermenunes09:
>
> What coincidence, I just started using this library today for a side project and needed the same.
> 

Thanks, @nearwood! It works great! I'm sure it will help lots of people who are trying to figure this out. I'm going to try to create a Polygon component based on that.

shane806:
[...]

PR #338, opened 2024-05-03
docs(examples): add 'Polyline' geometry component (#105)

disambiguationuk:
Addressing https://github.com/visgl/react-google-maps/issues/105

Added a new Polyline geometry component following the existing pattern with an example within the geometry app. Updated the Geometry readme.md to include the example. Basically the same as Polygon but with `encodedPath: string` instead of  `encodedPaths: string[]`

### DIFF
--- a/examples/geometry/README.md
+++ b/examples/geometry/README.md
@@ -130,3 +130,53 @@ interface PolygonProps extends google.maps.PolygonOptions {
 ```
 
 To see a Polygon on the Map, the `paths` or `encodedPaths` properties needs to be set.
+
+
+## `<Polyline>` Component
+
+React component to display a [Polyline](https://developers.google.com/maps/documentation/javascript/shapes#circles) instance.
+
+### Usage
+
+```tsx
+import React, {FunctionComponent} from 'react';
+import {APIProvider, Map} from '@vis.gl/react-google-maps';
+import {Polyline} from './components/polyline'; // import from your local file
+
+const App: FunctionComponent<Record<string, unknown>> = () => (
+  <APIProvider apiKey={'Your API key here'}>
+    <Map zoom={12} center={{lat: 53.54992, lng: 10.00678}}>
+      {/* Draw the Bermuda triangle */}
+      <Polyline
+        path={[
+          {lat: 25.774, lng: -80.19},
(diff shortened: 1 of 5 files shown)


disambiguationuk:
Fyi @Denness @joecarver @martinhale

usefulthink:
Awesome, thank you so much!

One more thing:

```
[warn] examples/geometry/README.md
[warn] examples/geometry/src/components/index.ts
[warn] Code style issues found in 2 files. Run Prettier to fix.
ERROR: "test:prettier" exited with 1.
```

examples/geometry/README.md
```diff
@@ -130,3 +130,53 @@ interface PolygonProps extends google.maps.PolygonOptions {
...
+### Usage
+
+```tsx
+import React, {FunctionComponent} from 'react';
+import {APIProvider, Map} from '@vis.gl/react-google-maps';
+import {Polyline} from './components/polyline'; // import from your local file
+
+const App: FunctionComponent<Record<string, unknown>> = () => (
+  <APIProvider apiKey={'Your API key here'}>
+    <Map zoom={12} center={{lat: 53.54992, lng: 10.00678}}>
```
usefulthink, line 148:
```suggestion
    <Map defaultZoom={5} defaultCenter={{lat: 24, lng: -72}}>
```

examples/geometry/src/components/polyline.tsx
```diff
@@ -0,0 +1,131 @@
...
+    onClick,
+    onDrag,
+    onDragStart,
+    onDragEnd,
+    onMouseOver,
+    onMouseOut,
+    encodedPath,
+    ...polylineOptions
+  } = props;
+  // This is here to avoid triggering the useEffect below when the callbacks change (which happen if the user didn't memoize them)
```
usefulthink, line 48: I like this technique, never thought about it this way. 
It's technically iccorrect, but works better than the default for 95% of use-cases.

usefulthink (approved):

examples/geometry/src/components/polyline.tsx
disambiguationuk: I just copied this from the Polygon example 😅

PR #339, opened 2024-05-03
chore(eslint): add useDeepCompareEffect to exhaustive deps rule

rodrigofariow:
eslint should now report missing dependencies when using `useDeepCompareEffect`.


usefulthink:
Good point, thanks! I was worried for a sec that I missed something 🙈

PR #351, opened 2024-05-09
feat(advanced-marker): remove content element in cleanup

Juneezee:
In PR https://github.com/visgl/react-google-maps/pull/335, the `InfoWindow` cleanup function removes the content element. We should implement the same behavior for the `AdvancedMarker` component to ensure proper cleanup of its content element as well.

https://github.com/visgl/react-google-maps/blob/bc10a6f3a64e7445bc8b672375c9d196b7899d9e/src/components/info-window.tsx#L78-L83


usefulthink:
Good point, thanks! I believe the AdvancedMarker should do it anyway, but this way it's clear and we don't have to "believe" :)

PR #363, opened 2024-05-11
added ControlPanel to marker-clustering example

luiscrjunior:
Just added ControlPanel to marker-clustering example.

The component was there, but it was not loded from `<App />`.


usefulthink:
Thanks!

Issue #386, opened 2024-05-28
[Bug] InfoWindow anchor reference not cleared on unmount


nickrschnelle:
### Description

When used with `@googlemaps/markerclusterer`, `InfoWindow`s do not seem to correctly clear their anchor reference during component dismount. This causes empty `InfoWindow`s to reappear against markers that have previously had their anchored `InfoWindow` closed.

<img width="520" alt="image" src="https://github.com/visgl/react-google-maps/assets/10275911/0f2e2e05-cac9-40d2-9816-30fad5db463c">

This issue does not seem to occur when you actually click the 'X' to close the `InfoWindow`. Only when clicking from one marker with an open `InfoWindow` to another marker and thus opening it's `InfoWindow` and un-mounting the first marker's `InfoWindow`

I'm not sure if this is best to post as part of the description however I was able to do some digging to diagnose the issue.
I modified the library source file `info-window.tsx` to add each created `google.maps.InfoWindow` into a global array which I then `console.log`ed out during the cleanup function of the `useEffect`

```
  // top of file
  const arrayOfWindows = [];
  
  ...

  useEffect(
    () => {
      if (!mapsLibrary) return;

      if (pixelOffset) {
        (infoWindowOptions as google.maps.InfoWindowOptions).pixelOffset =
          new google.maps.Size(pixelOffset[0], pixelOffset[1]);
      }

      // intentionally shadowing the state variables here
      const infoWindow = new google.maps.InfoWindow(infoWindowOptions);
>>    arrayOfWindows.push(infoWindow);
      const contentContainer = document.createElement('div');

      infoWindow.setContent(contentContainer);

      setInfoWindow(infoWindow);
      setContentContainer(contentContainer);

      // cleanup: remove infoWindow, all event listeners and contentElement
      return () => {
        console.log('infowindow cleanup lib');
        google.maps.event.clearInstanceListeners(infoWindow);
        infoWindow.close();
        contentContainer.remove();

        setInfoWindow(null);
        setContentContainer(null);
        // @ts-ignore
 >>     console.log('cleanup', arrayOfWindows);
      };
    }, [mapsLibrary]
  );
```

When reproducing the bug by opening an `InfoWindow` against one marker and then clicking another marker to open it's `InfoWindow` I was able to see the following logged out:
<img width="457" alt="image" src="https://github.com/visgl/react-google-maps/assets/10275911/d88989fd-dc44-4be2-8b8a-986115e1a316">

[...]

PR #390, opened 2024-05-29
fix: set infowindow anchor to null during component unmount

nickrschnelle:
fix for https://github.com/visgl/react-google-maps/issues/386

I had to do a bit of type coercion as `@types/google.maps` doesn't actually expose all the properties on the resulting object when you create a new `google.maps.InfoWindow`

### DIFF
--- a/src/components/info-window.tsx
+++ b/src/components/info-window.tsx
@@ -72,6 +72,9 @@ export const InfoWindow = (props: PropsWithChildren<InfoWindowProps>) => {
       }
 
       // intentionally shadowing the state variables here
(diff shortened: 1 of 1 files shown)


usefulthink:
I'll close this in favor of #393, I'll make sure to list you as Co-Author there :) 
Thanks for your help figuring this out!

nickrschnelle:
> I'll close this in favor of #393, I'll make sure to list you as Co-Author there :) Thanks for your help figuring this out!

thanks for fixing this so fast 😀

Issue #391, opened 2024-05-30
[Feat] Add padding option to defaultBounds


francescocretti:
### Target Use Case

Initialise the Map instance with default bounds and a padding value in pixels, without the need to call `fitBounds` right after initialising the map to apply padding.

### Proposal

The `defaultBounds` is very useful in several use cases to initialise a map with a specific view previously calculated from some incoming data (or any other source).
The bounds approach is also very convenient when it comes to padding: the Google Maps API accepts a padding in pixels as an argument of the `fitBounds` method ([docs here](https://developers.google.com/maps/documentation/javascript/reference/map#Map.fitBounds)).

Since defining the `defaultBounds` prop leads to invoking the `fitBounds` method under the hood (https://github.com/visgl/react-google-maps/blob/39603438948c285e8a14e7e4e8d01034b58e1db4/src/components/map/use-map-instance.ts#L142), it would be very handy to add the chance to pass also a padding option. 

Could be something like that:

```javascript
 <Map
   defaultBounds={bounds}
   defaultBoundsPadding={{
    left: 10,
    right: 10,
    top: 10,
    bottom: 10,
   }}
/>
```

It shouldn't be a big change, I'm available to help if needed.
Thanks a lot.

usefulthink:
I like this, in fact I haven't thought about that padding parameter. 

Maybe instead of adding an additional prop for that, we could add an optional field `padding` to the `defaultBounds` prop.

francescocretti:
@usefulthink sure, I don't have any preferences!
Using of an additional prop can be easier to handle and also guarantees backwards compatibility (since if u don't need padding you only pass `defaultBounds` as it is now).

Let me know what u prefer and I can open a PR if you'd like :)

usefulthink:
I think specifying the padding with the same prop you're using to specify the bounds makes sense, with padding being optional it also should be backwards-compatible. We could just extend the type in https://github.com/visgl/react-google-maps/blob/39603438948c285e8a14e7e4e8d01034b58e1db4/src/components/map/index.tsx#L82 to `google.maps.LatLngBoundsLiteral & {padding ?: number | google.maps.Padding}`.

francescocretti:
Okay, let's do it ;)

usefulthink:
This will be released with [v1.1.0](#395), probably around next week.

karldanninger:
I think this isn't working anymore.

```
const offset = { left: 40, right: 40, bottom: 40, top: 40 };
...
[...]

PR #392, opened 2024-05-30
Add padding option to `defaultBounds`

francescocretti:
As discussed in issue #391, this PR adds the optional `padding` property to `defaultBounds`.

e.g.

```javascript
    <Map
      defaultBounds={{
        west: -7.521054546833144,
        north: 57.18946776545227,
        south: 49.60181300450985,
        east: -2.8939254862738877,
        padding: {
          bottom: 50,
          right: 50,
        }
      }}
    />
```

NOTE
The Google Maps API implement a system of discreet zoom levels, so some values of padding could not be visible if they don't reach the smallest amount of padding necessary to cause the zoom level to change by a full integer value. More info: https://stackoverflow.com/a/44476181

### DIFF
--- a/docs/api-reference/components/map.md
+++ b/docs/api-reference/components/map.md
@@ -214,10 +214,16 @@ The initial state of the camera. This can be used to leave the map
 component in uncontrolled mode. When both a default-value and a controlled
 value are present for a parameter, the controlled value takes precedence.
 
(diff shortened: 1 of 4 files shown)


usefulthink:
That was fast, thanks a lot! I'll merge that later today for the 1.1.0 version, I still have a fix for 1.0.2 to do before that..

francescocretti:
@usefulthink no problem ;) thanks for your work. Let me know if changes are needed!

PR #428, opened 2024-06-23
Fix pin element overriding styling of advancedMarker

JCM00N:
Solves [this](https://github.com/visgl/react-google-maps/discussions/368#discussioncomment-9811113) as requested by 
@usefulthink


usefulthink:
this should be fixed with #472

PR #459, opened 2024-07-28
docs(examples): show more example tiles on small screen

shuuji3:
Hello, this PR is an adjustment to the stylesheet for the example page so that more example tiles can be shown on the small screen, by shrinking extra margin/padding and adjusting the width of a tile.

Currently, even a relatively large-sized phone can only show only one thumbnail in a row.

## Before
Screenshot from 2024-07-28 18-26-42
Screenshot from 2024-07-28 18-26-59

## After
Screenshot from 2024-07-28 18-26-50
Screenshot from 2024-07-28 18-27-06

## Desktop view
The desktop view remains the same.

### Before
Screenshot from 2024-07-28 18-29-30

### After
Screenshot from 2024-07-28 18-29-32

### DIFF
--- a/website/src/components/example/styled.js
+++ b/website/src/components/example/styled.js
@@ -8,6 +8,9 @@ export const ExampleHeader = styled.div`
   border-bottom: 1px solid 20px;
   display: inline-block;
   padding: 20px 20px 4px 0;
+  ${isMobile} {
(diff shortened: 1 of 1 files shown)


usefulthink:
Thank you very much for this!

Issue #299, opened 2024-04-07
`google.maps.Marker` deprecated


D3strukt0r:
### Description

Console throws following Warning when (I guess) using the `<Marker />` component

```
As of February 21st, 2024, google.maps.Marker is deprecated. Please use google.maps.marker.AdvancedMarkerElement instead. At this time, google.maps.Marker is not scheduled to be discontinued, but google.maps.marker.AdvancedMarkerElement is recommended over google.maps.Marker. While google.maps.Marker will continue to receive bug fixes for any major regressions, existing bugs in google.maps.Marker will not be addressed. At least 12 months notice will be given before support is discontinued. Please see https://developers.google.com/maps/deprecations for additional details and https://developers.google.com/maps/documentation/javascript/advanced-markers/migration for the migration guide.
```

### Steps to Reproduce

used a simple example

```js
<Map center={position} zoom={17}>
  <Marker position={position} />
</Map>
```

### Environment

- Library version: @vis.gl/react-google-maps@0.8.3
- Google maps version: weekly
- Browser and Version: Brave v1.64.116 (Chromium: v123.0.6312.105)
- OS: MacOS 14.4.1

usefulthink:
There's not really anything we can do about that, `google.maps.Marker` is officially deprecated and using it will trigger that warning-message. It's not going to go anywhere for the time being, but we recommend to use the `AdvancedMarker` component instead if possible.

D3strukt0r:
Oh didn't know about that component. completely new to this library. It's just that Marker is documented in basically all examples I've seen so far. Will update my code then :)

usefulthink:
Thanks for the hint, we'll review it and add notes to the docs and examples.

BParvatkar:
Can we update the doc JSDoc [here](https://github.com/visgl/react-google-maps/blob/f38d3c4004663fd1850c00dca7ddfb7e92b8d5cf/src/components/marker.tsx#L119) with a suggestion to use AdvancedMarker instead? 
Also, should we mark it as deprecated using JSDoc too? @usefulthink

PR #464, opened 2024-08-04
Fix Marker deprecation in  README.md

morgler:
Replace deprecated Marker with AdvancedMarker. Also adding a mapId for AdvancedMarker to work. Partially fixing #299


usefulthink (approved):

usefulthink:
Good point, thanks!

PR #472, opened 2024-08-06
feat: add hover events and anchor points to advanced markers

mrMetalWood:


usefulthink (commented):
Excellent, thanks a lot!

I think there's just two things missing here: updated documentation and some more explanations about the DOM structure being built for the marker.

What do you think, do we need to be worried that the added dom-element will cause backwards compatibility issues?

src/components/pin.tsx
```diff
@@ -58,7 +58,15 @@ export const Pin = (props: PropsWithChildren<PinProps>) => {
     }
 
     // Set content of Advanced Marker View to the Pin View element
-    advancedMarker.content = pinElement.element;
+    const markerContent = advancedMarker.content?.firstChild;
```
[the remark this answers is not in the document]
mrMetalWood: This fixes an issue that has already been addressed in this #428 where the the Pin component would overwrite the styles on an advanced marker by setting the content rather than replace the children.

 Below is just a more verbose way of saying `...replaceChildren(pinElement.element);` since that doesn't seem to be supported by our transpile target I think.

mrMetalWood:
Here is a comparison of the DOM structure output:

Before:
Screenshot 2024-09-05 at 12 17 47


After:
Screenshot 2024-09-05 at 11 57 50

src/components/info-window.tsx
```diff
@@ -176,6 +177,47 @@ export const InfoWindow = (props: PropsWithChildren<InfoWindowProps>) => {
...
+          anchor.content.getBoundingClientRect() ?? {};
+
+        // This checks whether or not the anchor has custom content with our own
+        // div wrapper. If not, that means we have a regular AdvancedMarker without any children.
+        // In that case we do not want to adjust the infowindow since it is all handled correctly
+        // by the Google Maps API.
+        if (anchorWidth === 0 && anchorHeight === 0) {
+          // We can safely typecast here since we control that element and we know that
+          // it is a div
+          const anchorDomContent = anchor.content.firstChild as Element;
```
usefulthink, line 193: This could be risky, I think `firstChild` could also be a text- or even a comment-node, using `Element.firstElementChild` should be fine here.

src/components/info-window.tsx
```diff
@@ -176,6 +177,47 @@ export const InfoWindow = (props: PropsWithChildren<InfoWindowProps>) => {
...
+          const anchorDomWidth =
+            anchorDomContent?.getBoundingClientRect().width;
+
+          const anchorComputedStyles = getComputedStyle(anchorDomContent);
+          const transformMatrix = new WebKitCSSMatrix(
```
usefulthink, line 198: Not sure about the `Webkit` prefix, I think it's now called [`DOMMatrix`](https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrix). 

If I understood this right, we need to compute the offset of the top/center of the marker-content to the current position of the marker. According to [the spec](https://drafts.csswg.org/cssom-view/#dom-element-getclientrects), getBoundingClientRect() already have all transforms applied, so we could maybe just take the values from there? Maybe the BCR of our 0x0 anchoring div could be used as reference to get to the actual offset.

usefulthink (commented):
very cool, almost there :)

usefulthink:
can you also have a look at this comment here: https://github.com/visgl/react-google-maps/discussions/368#discussioncomment-9806419? Is this going to be obsolete with the new implementation?

src/components/info-window.tsx
mrMetalWood: Good point :) Yes we can.

mrMetalWood:
> can you also have a look at this comment here: [#368 (reply in thread)](https://github.com/visgl/react-google-maps/discussions/368#discussioncomment-9806419)? Is this going to be obsolete with the new implementation?

Yes, I think that should be resolved with the new implementation

isaacpresidium:
> > can you also have a look at this comment here: [#368 (reply in thread)](https://github.com/visgl/react-google-maps/discussions/368#discussioncomment-9806419)? Is this going to be obsolete with the new implementation?
> 
> Yes, I think that should be resolved with the new implementation

Doesn't seem to be resolved for me, on latest version

PR #484, opened 2024-08-18
feat: add overlay view

itayphytech:


usefulthink:
Hey there, thanks for taking the time to contribute to this project! 

I noticed that you didn't provide any comments or explanation with this PR and it came sort of "out of the blue" (meaning, not referencing any previous discussion or similar). Can you share a bit about the motivation and intended uses for this component? What could be done with this implementation that can't already be done using the `AdvancedMarker`?

I already thought a bit about having an implementation of an OverlayView component, but I didn't yet end up at an idea that was versatile enough to cover all the possible uses of the OverlayView in a good way. 
Also, this topic ties into all the other types of overlays the Maps JavaScript API provides, and I would like to work on a good Vision for those as well before settling on a specific implementation.

Implementing this as an example first would be an excellent start. Besides documenting the intended usage of the component, this would also allow us to iterate on the idea and shape it into something that could be added to the library at some point.

Also, please take the time to review our [contribution guidelines](https://visgl.github.io/react-google-maps/docs/contributing) to understand where I'm coming from with this.

PR #541, opened 2024-09-19
feat: "onError" prop for api provider

### LINKED ISSUE #533: [Feat] Can we have an "onError" next to "onLoad"
### Target Use Case

Currently if we bump into https://github.com/visgl/react-google-maps/blob/main/src/components/api-provider.tsx#L178 then there's no way for consumer to retry or react; can we have an `onError` callback at this line?

### Proposal

```tsx

https://github.com/visgl/react-google-maps/blob/main/src/components/api-provider.tsx#L178

        } catch (error) {
          console.error(
            '<ApiProvider> failed to load the Google Maps JavaScript API',
            error
          );
          if (onError) {
            onError(error)
          }
        }
```

timmchi:
Resolves #533 

## Description

- Added onError to the list of API Provider props which will be called in case of an error when loading the Maps API. 
- Updated the mock in API Provider test to be able to simulate an error
- Added a unit test for the new functionality
- Added "Error Handling" section to the API Provider docs which mentions the new onError prop

I added the old console.error part to the else block so that there would be no error message in the console when running tests.

### DIFF
--- a/src/components/__tests__/api-provider.test.tsx
+++ b/src/components/__tests__/api-provider.test.tsx
@@ -27,8 +27,7 @@ const ContextSpyComponent = () => {
 ContextSpyComponent.spy = jest.fn();
 
 let triggerMapsApiLoaded: () => void;
-
-let shouldRejectLoad = false; // Variable to control the behavior of the mock. Used in test for onError callback
+let triggerLoadingFailed: () => void;
 
 jest.mock('../../libraries/google-maps-api-loader', () => {
   class GoogleMapsApiLoader {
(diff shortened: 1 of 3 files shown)


jtoloui:
Thanks @timmchi for this, I was having to add my own custom wrapper over the map to override the api loading which was very hacky,  do we know the expected timeline for this to be added and released?

usefulthink:
Sometime this week would be my guess :D

jtoloui:
Amazing cheers @usefulthink

src/components/__tests__/api-provider.test.tsx
```diff
@@ -38,14 +40,19 @@ jest.mock('../../libraries/google-maps-api-loader', () => {
...
 
       google.maps.importLibrary = importLibraryMock;
-      return new Promise(
-        resolve =>
-          (triggerMapsApiLoaded = () => {
+
+      return new Promise((resolve, reject) => {
+        if (shouldRejectLoad) {
+          reject(new Error('Test error: Failed to load Google Maps API'));
+        } else {
```
usefulthink, line 47: I would implement this the same way as the success-case is done, so, for example:

```tsx
triggerLoadingFailed = () => { 
  reject();
  onLoadingStatusChange(APILoadingStatus.FAILED);
};
```

then we get the same async behaviour as in the success tests.

usefulthink (changes requested):
Pretty cool, thanks a lot for the help with this!
I added a few small comments, let me know if you have any questions about that.

usefulthink (approved):

PR #551, opened 2024-09-27
[Bug 550] Optimized loadedLibraries reducer

jazzy-em:
Prevented unnecessary updating of context during AdvancedMarkers rendering


usefulthink (approved):

usefulthink:
Awesome, thanks a lot for spotting and fixing this. I wasn't even aware that there was a significant difference in the code-path taken between the `render` and `createRoot` approaches.

Issue #556, opened 2024-10-01
[Bug] Google Maps Marker Deprecation warning: "Google Maps Marker is deprecated, use AdvancedMarkerElement."


MohammadGhorayshi:
### Description

I have this warning in my browser console:
`As of February 21st, 2024 (v3.56), google.maps.Marker is deprecated. We encourage you to transition to the new google.maps.marker.AdvancedMarkerElement class. Advanced markers provide substantial improvements over the legacy google.maps.Marker class.`

I need to remove this warning from browser console and I only replace one line in your source code and work without any warning.

Change `google.maps.Marker` to `google.maps.marker.AdvancedMarkerElement`

### Steps to Reproduce

You need to use <Marker /> component and see this warning.

### Environment

- Library version: 1.3
- Google maps version: weekly
- Browser and Version: Chrome Version 129.0.6668.60 (Official Build) (arm64)
- OS: Mac

usefulthink:
Alternative: you can also change `import {Marker} from '@vis.gl/react-google-maps';` to `import {AdvancedMarker} from '@vis.gl/react-google-maps';` in your code and not use the legacy markers at all, as indicated by the deprecation-warning.

PR #557, opened 2024-10-01
fix: (issue-556) Google Maps Marker Deprecation warning

MohammadGhorayshi:
I have resolve this issue: https://github.com/visgl/react-google-maps/issues/556

### DIFF
--- a/src/components/marker.tsx
+++ b/src/components/marker.tsx
@@ -50,7 +50,7 @@ function useMarker(props: MarkerProps) {
       return;
     }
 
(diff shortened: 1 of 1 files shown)


usefulthink:
Sorry, but no. 

We already have the `AdvancedMarker` component if you don't need to use legacy markers. 
The `Marker` component exists to support legacy use-cases, since the `AdvancedMarkerElement` can't be used in all situations. We can't do anything about the deprecation notice, neither should we.

PR #567, opened 2024-10-10
Update README.md

supriya-thukral:
Updated yarn command


usefulthink:
In which situation are the quotes required?

Issue #569, opened 2024-10-13
[Bug] collisionBehavior, for whatever reason, isn't binding properly to the underlying AdvancedMarkerElement


bishoym:
### Description

Every other prop seems to be binding properly, except collision behaviour. 

Bizarre.

### Steps to Reproduce

Here's the codesandbox demonstrating the problem: https://codesandbox.io/p/sandbox/google-maps-collision-marker-test-2fy8lg

You can see that creating the markers and binding collisionBehavior imperatively makes it work, while the wrapped React component doesn't.

### Environment

- Library version: @vis.gl/react-google-maps@1.3.0
- Google maps version: weekly
- Browser and Version: Chromium (Arc v129.0.6668.90)
- OS: MacOS Version 14.6.1

usefulthink:
The first thing I noticed is that the property is named `collisionBehavior`, not `collisionBehaviour` (like the Maps JavaScript API, we use American English spelling in this library). Still, even with the correct name, it isn't working on the first render, which I will have to investigate.

mrMetalWood:
It seems like this bug got introduced by the new anchor handling of our advanced markers. Your demo is working when using the `1.1.3` version of the library. That might be a quick fix for you until we can figure out how to best address this.

bishoym:
Hey just wanted to give an update on this. I'm using the imperative approach until a fix is merged as that doesn't seem to have any downsides for my current use-case.

In the interim, if you think this is a good first issue I can try and jump in on the PR fixing this.

mrMetalWood:
@bishoym We published `1.4.0` which should fix the issue. Can you check if this solves your issues?

Issue #560, opened 2024-10-02
[Bug] AdvancedMarker custom icon not appearing


usefulthink:
### Discussed in https://github.com/visgl/react-google-maps/discussions/558

<div type='discussions-op-text'>

<sup>Originally posted by **Remarketeer** October  2, 2024</sup>
I'd like to ask for some help!

So a week ago I left my maps page alone to work on other stuff, but suddenly my markers disappeared. For some context, I have an array that I load information from and use .map to create markers for each entry. After some trial and error, I realized that it just wasn't accepting image inputs, e.g <img src='' width=.../>. 

I've attempted to load the image on my other pages and that was fine. I've attempted to load Pin objects and that worked too, I've tried loading just a singular marker using an image source but it behaves the same way. Loading the image directly on the maps page also seems to work so the issue shouldn't be that.  I'm quite stumped as to where it could have gone wrong.

Currently the data seems to have been loaded into the map as I've printed it to the console, but the marker just isn't appearing.</div>

mrMetalWood:
@Remarketeer I found an issue that would prevent a marker from showing when its width was set via CSS with a class on the `AdvancedMarker` element, but I'm not sure if that fits the issue you were describing. From the snippet you provided in the discussion it does not seem like it.

Could you provide a basic running example (on codesandbox or similar) where the issue exists?

usefulthink:
@Remarketeer with the new version 1.4.0, this should be fixed. Can you re-check if you're still seeing the problem?

Issue #559, opened 2024-10-02
[Bug] InfoWindow anchor not working


dante-blitz:
### Description

I have set an anchor on InfoWindow as per the docs...

```
        <AdvancedMarker
          position={center}
          onClick={toggleInfoWindow}
          ref={markerRef}
        >
          <Pin
            scale={2}
            background="#ff9200"
            borderColor="#000000"
            glyphColor="#000000"
          />
        </AdvancedMarker>
        {infowindowShown && (
          <InfoWindow anchor={marker} onCloseClick={closeInfoWindow}>
            {infoWindowContent}
          </InfoWindow>
        )}
```

The InfoWindow is placed directly on top on the Pin (z axis) and completely obscures it. This same code was previously working. Whenevener next.js hot-reloads, it appears to sort itself out and the InfoWindow is shifted up into the correct position.

### Steps to Reproduce

Load a page that has a gmap on it.

### Environment

"@vis.gl/react-google-maps": "^1.2.0",
"next": "14.2.14",

georgekolasa:
Having this exact issue as well.

dante-blitz:
Still have the issue as well. Haven't yet figured out a workaround that works for me. Manual pixel offset break on page reload or hydration when the anchor starts to work.

georgekolasa:
I'm not using NextJS personally.

dante-blitz:
@mrMetalWood This is happening in dev and production.

Tataro:
I have the same problem.

"next": "14.2.14",
"@vis.gl/react-google-maps": "^1.3.0"

mrMetalWood:
I was able to reproduce the bug and I am working on a fix.

mrMetalWood:
@dante-blitz @georgekolasa @Tataro We published `1.4.0` which should fix this issue. Can you check and see if that is the case for you? Would appreciate a short feedback if the issue is solved.

georgekolasa:
@mrMetalWood LGTM so far, thank you!

Issue #542, opened 2024-09-20
[Bug] onMouseEnter on AdvancedMarker only works when onClick is set


antonio-pbilby:
### Description

Using the onMouseEnter with a AdvancedMarker is only working when the same AdvancedMarker has a onClick property set.
If I'm using the standar Pin, it works just fine.

### Steps to Reproduce

```tsx
<APIProvider
  apiKey={import.meta.env.VITE_API_GOOGLE_MAPS_API_KEY}
  region="BR"
  language="pt-BR"

>
  <Map
    defaultCenter={{
      lat: -15.4,
      lng: -47.4,
    }}
    defaultZoom={10}
    id="plots-list"
    mapId={import.meta.env.VITE_API_GOOGLE_MAPS_MAP_ID}
  >
    {plotLocations.map((plot) => {
      return <AdvancedMarker position={plot.location} key={plot.id} onMouseEnter={() => console.log('test')}
        onClick={() => {}}
      >
        <div>dummy component</div>
      </AdvancedMarker>
    })}
  </Map>
</APIProvider>
```

### Environment

- Library version: 1.2.0
- Google maps version: weekly
- Browser and Version: Chrome 128.0.6613.138 (Official Build) (64-bit)
- OS: Windows 11

usefulthink:
@mrMetalWood can you have a look at that? This sounds like it could have something to do with the `gmpClickable` option for the advanced markers, which will be set when an onClick callback is defined. 

@antonio-pbilby can you check if it works for you when you manually add the `clickable` prop to your markers?

https://github.com/visgl/react-google-maps/blob/43b5c95c7e87443c49d687b075b688e884d925b1/src/components/advanced-marker.tsx#L236-L244

antonio-pbilby:
> @mrMetalWood can you have a look at that? This sounds like it could have something to do with the `gmpClickable` option for the advanced markers, which will be set when an onClick callback is defined.
> 
> @antonio-pbilby can you check if it works for you when you manually add the `clickable` prop to your markers?
> 
> https://github.com/visgl/react-google-maps/blob/43b5c95c7e87443c49d687b075b688e884d925b1/src/components/advanced-marker.tsx#L236-L244

It worked with the default Pin, but not with my custom component.

Does not work:
```tsx
<AdvancedMarker position={plot.location} key={plot.id}
  onMouseEnter={() => setHighlightedPlot(plot.id)}
  onMouseLeave={() => setHighlightedPlot(null)}
  clickable
[...]

PR #577, opened 2024-10-17
fix: advanced marker anchoring

mrMetalWood:
First draft for improved and more robust advanced marker anchoring. Still needs some testing...

Main change is moving away from the "0, 0" div for the anchoring reset. That seemed to cause problems in a few situations

This should address and fix:
- #569 
- #559
- #542 
- #560

### DIFF
--- a/src/components/__tests__/advanced-marker.test.tsx
+++ b/src/components/__tests__/advanced-marker.test.tsx
@@ -148,8 +148,10 @@ describe('map and marker-library loaded', () => {
       .get(google.maps.marker.AdvancedMarkerElement)
       .at(0) as google.maps.marker.AdvancedMarkerElement;
 
-    expect(marker.content?.firstChild).toHaveClass('classname-test');
-    expect(marker.content?.firstChild).toHaveStyle('width: 200px');
+    expect(marker.content?.firstChild?.firstChild).toHaveClass(
+      'classname-test'
+    );
+    expect(marker.content?.firstChild?.firstChild).toHaveStyle('width: 200px');
     expect(
       queryByTestId(marker.content as HTMLElement, 'marker-content')
     ).toBeTruthy();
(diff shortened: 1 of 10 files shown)


src/components/__tests__/advanced-marker.test.tsx
```diff
@@ -148,8 +148,10 @@ describe('map and marker-library loaded', () => {
       .get(google.maps.marker.AdvancedMarkerElement)
       .at(0) as google.maps.marker.AdvancedMarkerElement;
 
-    expect(marker.content?.firstChild).toHaveClass('classname-test');
-    expect(marker.content?.firstChild).toHaveStyle('width: 200px');
+    expect(marker.content?.firstChild?.firstChild).toHaveClass(
```
usefulthink, line 151: wondering if we should replace this with something like "expect that there is an element with classname-test somewhere in marker.content" – feels like specifying the exact hierarchy here isn't actually what we want to test.

src/components/advanced-marker.tsx
```diff
@@ -31,6 +31,10 @@ export function isAdvancedMarker(
   );
 }
 
+function isElementNode(node: Node): node is HTMLElement {
+  return (node as Node).nodeType === Node.ELEMENT_NODE;
```
usefulthink, line 35: isn't `as Node` redundant here, since that's the type given above?

usefulthink (commented):
Looking really good! 

What I would really like to have still is an overview of which elements are created, when and why, and how this all works together. Not sure if that has to be in the documentation or just somewhere in the code. We could also tackle this in a follow-up.

src/components/advanced-marker.tsx
mrMetalWood: Ah yes, you're right.

src/components/__tests__/advanced-marker.test.tsx
mrMetalWood: Yes, that sounds like a better approach.

mrMetalWood:
> Looking really good!
> 
> What I would really like to have still is an overview of which elements are created, when and why, and how this all works together. Not sure if that has to be in the documentation or just somewhere in the code. We could also tackle this in a follow-up.

Agreed, that would be helpful.

usefulthink (approved):

Issue #480, opened 2024-08-12
[Feat] Static fallback


housseindjirdeh:
### Target Use Case

Minimize the user experience impact of the library by improving paint time and responsiveness metrics.

### Proposal

Data from HTTP Archive shows that Google Maps has the longest download times during page rendering when compared to the other most popular third-party libraries on the web:

| <img src="https://github.com/user-attachments/assets/fe45f3fa-44cc-45d4-aeed-f34e8c842c14" width=80%> | 
|:--:| 
| Median download time during page rendering for the top 10 most popular third parties (calculated using HTTP Archive and Lighthouse’s wastedMs attribute via the render blocking audit) 
[Query](https://console.cloud.google.com/bigquery?sq=1018750324929:6e459f3c6a73471bae5f12ead005ea7c) |

This happens because the Maps JavaScript API fetches ~240 kB of JavaScript across multiple scripts. In this library, we can circumvent this by delaying the instantiation of these scripts until the user engages with the map. While waiting for the user’s interaction, we can fetch and display an image via the [Maps Static API](https://developers.google.com/maps/documentation/maps-static/overview) passing in the same set of required props (`center` and `zoom`).

| <img src="https://github.com/user-attachments/assets/b7f7dc2d-270f-475a-ae0d-4a5bd36d6572" width=80%> | 
|:--:| 
| Optimized prototype of `react-google-maps` that displays an image and defers JS until user interaction) 
[Link](https://react-google-maps-nextjs-optimized.vercel.app/) |

To implement this feature, the `<APIProvider>` component of the library can keep track of whether the map has been selected as part of the context information it provides. We can use this to conditionally render either the “real” map versus a static version of it:

```jsx
// src/components/map/index.tsx

import { StaticMap } from "./static-map";
// ...

return (
  <div onClick={() => context.setMapSelected(true)}>
    >
    {context.mapSelected ? (
      // Load normal map
    ) : (
      <StaticMap {...props} /> // Load Static Map if map hasn't been clicked
    )}
  </div>
);
```

You can see all the changes in this prototype in this [commit](https://github.com/visgl/react-google-maps/compare/main...housseindjirdeh:react-google-maps:static-facade). Note that this is just a prototype, and the real implementation would offer some sort of UI that makes it clear that the map needs to be clicked in order to be interacted with.

### Performance Tests

mrMetalWood:
Hi @housseindjirdeh,

wow 😍 what a detailed and thought out issue and proposal!

[...]

PR #579, opened 2024-10-17
Initial `<StaticMap>` implementation

housseindjirdeh:
A basic, initial implementation of a `StaticMap` component for #480. Tested locally on a Next.js site:

image

-----

### Handling Signatures

Note: This also supports signed URLs, but curious to know if this approach looks correct. Since this is all constructed client-side, I included an optional `signature` prop. For support in SSR frameworks like Next.js, we may want to include some documentation to show how to construct a signature server-side in a route handler or server action. 

For example, something like:

```js
// Next.js
// /app/api/signature.js

import { NextResponse } from 'next/server';
import { createSignature } from '@googlemaps/url-signature';

export async function POST(request) {
  try {
    const { mapOptions } = await request.json(); // Get map parameters from the request

    let pathAndQuery = '/maps/api/staticmap';
    pathAndQuery += new URLSearchParams(mapOptions).toString(); // Add map options

   const signature = createSignature(pathAndQuery, process.env.GOOGLE_MAPS_SIGNING_SECRET);

    return NextResponse.json({ signedUrl });
  } catch (error) {
    return NextResponse.json({ error: "Failed to generate signature" }, { status: 500 });
  }
}
```

And then the signature can be passed down client-side:

```jsx
'use client';

import { useState, useEffect } from 'react';
import { APIProvider, StaticMap } from 'react-google-maps';

export function GoogleMaps({ center, zoom, size, maptype }) {
  const [signature, setSignature] = useState(null);

  useEffect(() => {
    const getSignature = async () => {
      try {
        const mapOptions = {
          center: `${center.lat},${center.lng}`, 
          zoom,
          size,
          maptype
        };

        const response = await fetch('/api/signature', {
          method: 'POST',
          headers: { 'Content-Type': 'application/json' },
          body: JSON.stringify({ mapOptions })
        });

        const data = await response.json();
        setSignature(data.signature); 
      } catch (error) {
        console.error("Error fetching signature:", error);
      }
    };

    getSignature();
  }, [center, zoom, size, maptype]);

  if (!signature) {
    return <div>Loading map...</div>; 
  }

  return (
    <APIProvider apiKey={process.env.NEXT_PUBLIC_GOOGLE_MAPS_API_KEY}>
      <StaticMap 
        defaultCenter={center} 
        zoom={zoom} 
        size={size}
        maptype={maptype}
        signature={signature} 
      />
    </APIProvider>
  );
}
```

### DIFF
--- a/src/components/api-provider.tsx
+++ b/src/components/api-provider.tsx
@@ -23,6 +23,7 @@ export interface APIProviderContextValue {
   loadedLibraries: LoadedLibraries;
   importLibrary: typeof google.maps.importLibrary;
   mapInstances: Record<string, google.maps.Map>;
+  apiKey: string;
   addMapInstance: (map: google.maps.Map, id?: string) => void;
   removeMapInstance: (id?: string) => void;
   clearMapInstances: () => void;
@@ -230,7 +231,8 @@ export const APIProvider = (
       clearMapInstances,
       status,
       loadedLibraries,
-      importLibrary
+      importLibrary,
+      apiKey: loaderProps.apiKey
     }),
     [
       mapInstances,
@@ -239,7 +241,8 @@ export const APIProvider = (
       clearMapInstances,
       status,
       loadedLibraries,
-      importLibrary
+      importLibrary,
+      loaderProps.apiKey
     ]
   );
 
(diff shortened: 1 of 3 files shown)


usefulthink:
Hi @housseindjirdeh,

sorry for not getting back to you earlier. We took your work as a starting point, but eventually ended up taking this in a different direction. We wanted the component to be built in a way that allows server-side rendering of the `StaticMap` component, so we can get the most out of the possible performance gains. 

I'll be closing this now since the completed implementation was merged via #633. 

Thanks again for the initiative and the initial implementation!

Issue #576, opened 2024-10-17
[Feat] Support of channel prop in APIProvider component


yashSangai:
### Target Use Case

By addition of this, users would be able to track usage and bills as per channel.

Reference: https://developers.google.com/maps/reporting-and-monitoring/reporting#usage-tracking-per-channel

### Proposal

The code would look something like below if channel prop is added:

```
import React from 'react';
import {APIProvider} from '@vis.gl/react-google-maps';

const App = () => (
  <APIProvider apiKey={'Your API key here'} channel={<Your channel number>}>
    {/* ... any components ... */}
  </APIProvider>
);
export default App;
```

PR #584, opened 2024-10-22
Channel Prop Addition in APIProvider Component

yashSangai:
Implementation of this feature request: https://github.com/visgl/react-google-maps/issues/576


src/components/api-provider.tsx
```diff
@@ -74,6 +74,12 @@ export type APIProviderProps = {
    * empty string. Read more in the
    * [documentation](https://developers.google.com/maps/reporting-and-monitoring/reporting#solutions-usage).
    */
+  channel?: string;
```
usefulthink, line 77: If it's only numbers, should we maybe accept `string | number` as type here?

usefulthink (commented):
Awesome, thank you very much for this! Besides the one question this looks already good to go.

src/components/api-provider.tsx
yashSangai: @usefulthink Actually the urlParam expect key and value to be string because of which I changed the type to string BUT now I have got a solution for it so I changed the type back to number.

usefulthink (approved):

PR #633, opened 2024-12-09
feat: add static maps component with SSR support

mrMetalWood:
- Make the Static Maps Api (https://developers.google.com/maps/documentation/maps-static/start) available via a `StaticMap` component and a `createStaticMapsUrl` helper function. The helper functions main purpose is to help with server side "url signing" which could look very different depending on the individual setup.
- experiment with SSR support for the new static-map component. Export the `StaticMap` component from a new `@vis.gl/react-google-maps/server` endpoint which can be used for SSR. It is also exported from the main entry `@vis.gl/react-google-maps`


usefulthink (approved):
Awesome work, thanks! The only thing I would like to add is to make the StaticMap component a bit more useful – as it is now, it could just as well be removed.

Issue #628, opened 2024-12-04
[Bug] InfoWindow automatically closes when the map moves.


Bcavez:
### Description

On version 1.4.2, the InfoWindow automatically closes when the map is moved. Which is particularly annoying because when an InfoWindow is not entirely in the viewport, the map is automatically moved in order to fully show the infowindow, which means that the InfoWindow is opened then instantly closed.

Looking at it, I can see that the onClose event is triggered whenever the map is being moved.
I don't know if this is relevant, but I'm using the infowindow on an advancermarker and there are also clusters on the map.

On version 1.1.3, everything works fine. When opening the InfoWindow, it stays open. When moving the map, it stays open.

### Steps to Reproduce

First case:

1) Add an InfoWindow to an AdvancerMarker like described in the docs https://visgl.github.io/react-google-maps/docs/api-reference/components/info-window#infowindow-attached-to-marker
2) Click on the InfoWindow
3) Move the map, the window is closed.

Second case:

1) Add an InfoWindow to an AdvancerMarker like described in the docs https://visgl.github.io/react-google-maps/docs/api-reference/components/info-window#infowindow-attached-to-marker
2) Move the map so that the marker is in a corner and there won't be enough space to display the infowindow
3) Click on the marker
4) The infowindow opens then closes instantly after map re-centering

### Environment

- Library version: "@vis.gl/react-google-maps": "1.4.2"
- Google maps version: weekly
- Browser and Version: Chrome latest
- OS: windows 11

mrMetalWood:
Hi @Bcavez, thanks for taking the time to report the issue! It looks like a bug indeed. At first glance it seems that this effect (https://github.com/visgl/react-google-maps/blob/main/src/components/info-window.tsx#L175-L235) is triggered more often than it should and thus calling `infowindow.close()` which then triggers the `onClose` event. The fix should not be pretty straight forward. Maybe even just replacing `useEffect` with `useDeepCompareEffect`

Bcavez:
@usefulthink @mrMetalWood 

I think this issue should be re-opened as case N°1 has indeed been fixed, but not case N°2.

usefulthink:
[...]

PR #642, opened 2024-12-18
fix: use deep compare effect to prevent infowindow close

mrMetalWood:
The `infowindowOptions`  was a new object on every render thus resulting in the infowindow getting closed when the map was panned or zoomed

addresses #628


kylegoines:
hello! i was hoping to know if theres going to be a release for this fix, its vital to my project and the release schedule :)

PR #675, opened 2025-01-27
Improve map ColorScheme and RenderingType types

TiagoPortfolio:
Changed the ColorScheme and RenderingType types to use const to avoid being inferred as a string instead of the literal value.


usefulthink:
Thanks, that seems like a good idea! 
I don't think this should be a problem for any existing users, correct?

TiagoPortfolio:
> Thanks, that seems like a good idea! I don't think this should be a problem for any existing users, correct?

Correct.
If anything, existing users will be warned if they are using an invalid color scheme or rendering type.

PR #682, opened 2025-02-07
chore: add internalUsageAttributionIds in the library

zhunhan:
This PR add internalUsageAttributionId to help Google understand which libraries and samples are helpful to developers, such as usage of this library. Instructions for opting out of the identifier are provided in README.

The value is assign as part of APIProvider context and passed to maps instance to pass to mapOptions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,17 @@ const App = () => {
   );
 };
 ```
+### Internal usage attribution ID
+
+This library uses [`internalUsageAttributionIds`](https://developers.google.com/maps/documentation/javascript/reference/map#MapOptions.internalUsageAttributionIds), which helps Google understand which libraries and
+samples are helpful to developers and is optional.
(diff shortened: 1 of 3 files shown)


usefulthink:
Thanks! Is this used as a replacement for what we're currently doing with the solution_channel parameter?

zhunhan:
> Thanks! Is this used as a replacement for what we're currently doing with the solution_channel parameter?

Yes, that's the plan, right now we will use both and migrate to internalUsageAttributionIds as the sole one in the future.

zhunhan:
Converting the PR to draft for now. 

The update requires Google Maps JS 3.59.2, whereas the current library only requires 3.54.10. Let me check offline on the version bump.

usefulthink:
I'll reopen this, will be a lot easier.

PR #683, opened 2025-02-07
[FEATURE] Draggable Directions example

### LINKED ISSUE #672: Draggable Directions example
### Target Use Case

Is it possible to get an example of how we could use this package with draggable directions?

### Proposal

Implementing something similar to [this](https://developers.google.com/maps/documentation/javascript/examples/directions-draggable) example

adaup1:
Implemented draggable map markers into the Directions example. Added comments into code to specify what is necessary for draggable directions. Mentioned feature in README. 

This will close #672 

draggable-directions-example

### DIFF
--- a/examples/directions/README.md
+++ b/examples/directions/README.md
@@ -6,6 +6,8 @@ This is an example which shows how to use `useMapsLibrary` to load the `routes` 
 
 It allows the user to choose alternative routes, updating the route being rendered on the map.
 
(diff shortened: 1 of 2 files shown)


usefulthink:
Awesome! Thank you very much for this.

raffidahmad:
is there a graceful way to handle errors when it returns DIRECTIONS_ROUTE: ZERO_RESULTS

PR #692, opened 2025-02-22
fix: 3d map not loading correctly due to new parameter mode=hybrid not included

afian:
On Fri 21 Feb, Google Maps released a breaking change that required mode="hybrid" added to the  <gmp-map-3d/> component. This change fixed that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /examples/**/node_modules
 /examples/**/dist
 /examples/**/.env*
+.vscode/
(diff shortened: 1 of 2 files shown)


shuuji3:
There seems to be another option `satellite`. How about adding a new prop named `mapMode` to make it selectable?

docs: https://developers.google.com/maps/documentation/javascript/reference/3d-map#Map3DElement.mode

afian:
Yes I can do that over the weekend - but please note that the 3D demo at https://visgl.github.io/react-google-maps/examples/map-3d is broken for now.

Screenshot 2025-02-21 at 8 03 31 PM

usefulthink:
fixes #693

usefulthink:
Thanks a lot for providing a fix so quickly – I didn't even notice from the announcement that after this change, the 3d map won't render if the mode attribute is missing.

If you want to create another PR to add a prop for the Map3D component, you can just use `mode` as prop-name, but let's make it optional (default should be `hybrid`).

PR #718, opened 2025-03-19
Doc(example): Migrate autocomplete-custom to new Places API

yannagrebah:
Greetings everyone,

Following my [discussion](https://github.com/visgl/react-google-maps/discussions/707) regarding using the new Places API. This is my attempt at migrating the Autocomplete examples, this part is for `autocomplete-custom`.

Functionally and visually, it works the same, with the added improvement of the new API of not having to use `autocompleteService`, `placesService` and `useMap`.

I have also made sure to keep the other components compatible until me or somebody else migrate them as well (although I have doubts with `classic` getting the treatment for the time being as it is still in experimental phase, but hope for the best!)

Obviously, you need to have the new Places API enabled in your project to be able to use this.


usefulthink:
I thought that as well – we should keep the old one around (probably renamed to `autocomplete-legacy` or something like that) and add a notice and link to the new example.  For now it is much better to keep going it like this, as it makes reviewing the changes a lot easier.

usefulthink:
I'm currently adding a couple of things to this PR – I want to get it merged as soon as possible.

PR #745, opened 2025-04-21
Update static-map.md small spelling error under StaticMapsPath

meints12:
Just fixing this typo from bei to be!


usefulthink:
thanks a lot!

PR #747, opened 2025-04-25
fix: ensure proper type inference of APILoadingStatus

NicoToff:
Added `as const` to ensure that `APILoadingStatus` is the intended union type instead of `string`


usefulthink:
Thanks a lot!

PR #754, opened 2025-05-01
refactor: no `Node` type casting

mxdvl:
## Why are we doing this?

Because there is a type predicate already, so we can [keep things consistent](https://github.com/visgl/react-google-maps/blob/42602d6c20d29b176c8e97ee568a9665aac69309/src/components/advanced-marker.tsx#L224).


usefulthink:
thanks!

Issue #563, opened 2024-10-05
[Bug] when zooming fast back and forth the map gets to a state that is stuck


ronlv10:
### Description

When using **a phone** and zooming fast back and forth (out in out) with a map with advanced markers + clustering, sometimes the map gets to a state that it's stuck (and "going crazy")

after it happens a user can't do anything on the map and sometimes it's stuck in a way that reloads the tab doesn't work.

See:

https://github.com/user-attachments/assets/d85c4741-13a3-4fef-908e-c5f353e2cf72

P.S It also happened in my implementation of clustering on a map with an app that I have built.

### Steps to Reproduce

not an easy one. I used the clustering demo : https://visgl.github.io/react-google-maps/examples/custom-marker-clustering

you should use your phone and zoom in and out aggressively 

### Environment

- Library version:  1.3.0
- Google maps version: weekly
- Browser and Version: chrome
- OS: android

### Logs

```text
there is an warning when it happens: 

Cannot remove a WebglOverlay that has not been set to a map
```
```

**Comment by ronlv10** [comment]
the error is the same as https://github.com/visgl/react-google-maps/issues/419 but in this bug it says that there is no impact on the app. Here the bug is critical.

**Comment by ronlv10** [comment]
hey, @usefulthink  any comment? 
This bug impacts our app severely

**Comment by usefulthink** [comment]
As far as I know, the issue in #419 has been resolved upstream in the maps API itself. Are you still seeing that warning-message?

**Comment by ronlv10** [comment]
> And you're saying it is severely affecting your app, can you elaborate on that a bit? I was under the impression that the original issue only happens occasionally when aggressively zooming around the screen, which isn't something I'd expect users to do.

It happened to me multiple times while doing usability session with users. It doesn't happen only when zooming aggressively but it was my way to reproduce it faster. 
Here are two videos that I made now, zooming slowly back and forth until it happens: one from the demo clustering and one from my app. (please note it happens only with mobile zoom (two fingers)

Demo:

https://github.com/user-attachments/assets/fe47c598-1404-45b5-a1ad-b5beff922bcc

My app:

https://github.com/user-attachments/assets/8b34b524-1f9f-4fda-b816-4db389423684



> As far as I know, the issue in #419 has been resolved upstream in the maps API itself. Are you still seeing that warning-message?

After upgrade the package version I didn't see the warning, I did saw now this error:
```
[...]

PR #806, opened 2025-07-28
fix: Update calc clusters to happen after Idle event - reduces issue 563

ronlv10:
### Summary

This PR updates the `custom-marker-clustering` example to delay cluster rendering until the map becomes idle, rather than reacting to every `bounds_changed` event during zoom.

### Motivation

While investigating [issue #563](https://github.com/visgl/react-google-maps/issues/563), I found that the aggressive re-clustering during zoom gestures may contribute to a rare but severe bug where the zoom level resets unexpectedly (sometimes to zero). This is especially noticeable with a large number of markers. See [comment](https://github.com/visgl/react-google-maps/issues/563#issuecomment-3128047435).

By aligning the behavior with how Google’s official `[js-markerclusterer](https://github.com/googlemaps/js-markerclusterer)`(https://github.com/googlemaps/js-markerclusterer) handles clustering (listening to the `idle` event), we reduce the likelihood of triggering this issue and improve performance during zoom gestures.

### DIFF
--- a/examples/custom-marker-clustering/src/hooks/use-map-viewport.ts
+++ b/examples/custom-marker-clustering/src/hooks/use-map-viewport.ts
@@ -15,7 +15,7 @@ export function useMapViewport({padding = 0}: MapViewportOptions = {}) {
   useEffect(() => {
     if (!map) return;
 
-    const listener = map.addListener('bounds_changed', () => {
+    const listener = map.addListener('idle', () => {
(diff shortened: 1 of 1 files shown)


usefulthink:
That seems reasonable. 
Although the clustering (or rather, retrieving the clusters) is pretty fast, it makes total sense to debounce that a bit, and waiting for the idle event is a great way to do that.

Issue #808, opened 2025-08-01
[Bug] Static Map merges coordinate paths with the same style


Mauriciocr207:
### Description

I expected that each path passed in the paths array would be drawn independently. Instead, the paths are grouped by their style. When multiple paths share the same style, their coordinate arrays are merged into a single path.

I checked the implementation in the file:

`src/libraries/create-static-maps-url/assemble-path-params.ts`

And it seems that grouping paths by style is the intended behavior, as seen in this function:

`src/libraries/create-static-maps-url/assemble-path-params.ts:`
```javascript
import {formatLocation, formatParam} from './helpers';
import {StaticMapsPath} from './types';

/**
 * Assembles path parameters for the Static Maps Api from an array of paths.
 *
 * This function groups paths by their style properties (color, weight, fillcolor, geodesic)
 * and then constructs a string of path parameters for each group. Each path parameter string
 * includes the style properties and the coordinates of the paths.
 *
 * @param {Array<StaticMapsPath>} [paths=[]] - An array of paths to be assembled into path parameters.
 * @returns {Array<string>} An array of path parameter strings.
 *
 * @example
 * const paths = [
 *   {
 *     color: 'red',
 *     weight: 5,
 *     coordinates: [
 *       { lat: 40.714728, lng: -73.998672 },
 *       { lat: 40.718217, lng: -73.998284 }
 *     ]
 *   }
 * ];
 *
 * const pathParams = assemblePathParams(paths);
 * Output: [
 *    'color:red|weight:5|40.714728,-73.998672|40.718217,-73.998284'
 *  ]
 */
export function assemblePathParams(paths: Array<StaticMapsPath> = []) {
  const pathParams: Array<string> = [];

  // Group paths by their style properties (color, weight, fillcolor, geodesic)
  // to combine paths with identical styles into single parameter strings
  const pathsByStyle = paths?.reduce(
    (styles, path) => {
      const {color = 'default', weight, fillcolor, geodesic} = path;

      // Create unique key for this style combination
      const key = [color, weight, fillcolor, geodesic]
        .filter(Boolean)
        .join('-');

      styles[key] = styles[key] || [];
      styles[key].push(path);
      return styles;
    },
    {} as Record<string, Array<StaticMapsPath>>
  );

  // Process each group of paths with identical styles
  Object.values(pathsByStyle ?? {}).forEach(paths => {
    let pathParam = '';

    // Build style parameter string using properties from first path in group
    // since all paths in this group share the same style
[...]

PR #809, opened 2025-08-07
fix: remove path grouping by style for static maps url

mrMetalWood:
Removes the paths grouping by style for the Static Maps url. Does not make sense to do that to save a few bytes. This caused issues when multiple paths with the same same styles need to be drawn that are not connected.

addresses #808


usefulthink:
Thanks a lot!

PR #866, opened 2025-11-05
refactor(AdvancedMarker): improve stability of position and event handlers

rodrigofariow:
Hi there 👋 
For a library like this one that relies heavily on correct and stable rendering of useEffect and useMemo dependencies, I suggest we add a useEffectEvent ponyfill, to fill that gap.
This ponyfill is proven to work and the new version of eslint react-hooks plugin, that is already installed, supports verifying and using these events.

I took the freedom to add it in key places like useMapsEventListener and useDomEventListener.

### DIFF
--- a/src/components/__tests__/info-window.test.tsx
+++ b/src/components/__tests__/info-window.test.tsx
@@ -274,24 +274,37 @@ describe('<InfoWindow> events', () => {
     const onCloseSpy = jest.fn();
     const onCloseClickSpy = jest.fn();
 
+    // stup the google.maps.event mock
+    const listeners: Record<string, () => void> = {};
     const gme = jest.mocked(google.maps.event);
-    gme.addListener.mockImplementation(() => ({remove: jest.fn()}));
+    gme.addListener.mockImplementation((_, name, handler) => {
+      listeners[name as string] = handler as () => void;
+      return {remove: jest.fn()};
+    });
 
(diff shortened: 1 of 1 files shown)


src/components/advanced-marker.tsx
```diff
@@ -200,7 +200,25 @@ export type CustomMarkerContent = HTMLDivElement | null;
...
+  collisionBehavior,
+  clickable,
+  draggable,
+  position: positionProp,
+  title,
+  zIndex,
+  anchorPoint,
+  anchorLeft,
+  anchorTop
+}: AdvancedMarkerProps) {
```
usefulthink: That's for sure a matter of taste, but please keep the destructuring in the method-body, I prefer having the signature compact, so it can be quickly read all at once.

src/hooks/useEffectEvent.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * This file is based on https://github.com/sanity-io/use-effect-event/blob/main/src/useEffectEvent.ts
+ */
```
usefulthink:
```suggestion
// Copyright (c) 2025 Sanity
// This file is derived from sanity-io/use-effect-event (Licensed under the MIT License)
//
// https://github.com/sanity-io/use-effect-event
```

Issue #832, opened 2025-09-22
[Bug] InfoWindow flickering on re render (creating several portals?)


hugoatmooven:
### Description

When I open an InfoWindow for the first time, it works fine. But after I update the position (I tested with position, but could be true for other props), I can see the content flickering every time it re renders. It seems to get worse with every position change. If the position changed once, the content will blink once on re render. If position changed twice, content will blink twice and so on.

Maybe it's adding a new portal with every position change? I managed to fix the problem in some cases by using position as key of InfoWindow, which could indicate that something is not being cleaned up properly on props update. The "key" trick doesn't work when headerContent is set.

### Steps to Reproduce

This is not the exact same scenario as I described, but it is an easier way to reproduce the same problem. In this case I am using the flickering to trigger an infinite loop: mouse enter -> update state ->  flicker ->  flicker causes mouse leave -> update state -> flicker -> mouse enter

CodeSandbox example: https://codesandbox.io/p/devbox/marker-issue-forked-kssrss?workspaceId=ws_AJApL5q2Q836HiYdKXszjW

- Click anywhere on the map to open the InfoWindow
- Notice that you can hover over the content elements to change the box color, at least the first time InfoWindow is opened
- Click on another location on the map to move the InfoWindow
- Hovering the elements should now trigger an infinite loop

### Environment

- Library version: 1.5.5
- Google maps version: weekly
- Browser and Version: Chrome Version 140.0.7339.133
- OS: MacOS 15.6

### Logs

```text

```

PR #903, opened 2026-01-27
Fix flickering of map components

hugoatmooven:
As described in [this issue](https://github.com/visgl/react-google-maps/issues/832), the info window will flicker under certain conditions when a parent re renders.
Since this issue was opened, I upgraded the library and noticed other components started flickering with no apparent reason as well. The Map component seems to be setting map options on every render, even if the options haven't changed. It's not obvious with default settings, but if you have custom options for UI controls, it becomes very apparent. In my case I have custom values for `mapTypeControlOptions`, and the map type buttons will flicker like crazy by simply panning the map.

**General issues**

I found the source of the problem in use-memoized hook. I think there is a fundamental mistake in the memoization logic, causing the value to update even when it hasn't changed. It uses use-previous hook to compare the current value with the previous value, then returns the previous value if no changes occurred. It works on the first re-render, but on the next render, what is now the current value, will become the previous value, so even without change it will return a different object than the original memoized value.
That problem went away when I reverted the change made by [this pr](https://github.com/visgl/react-google-maps/pull/851/changes) to the use-memoized hook.

**Info window issue**

Even after fixing the memoization, I still had re-render problems with the info window component.
I found a couple of effects in the component that were directly mutating the memoized value of the options object, causing it to fail the deep equal test every time. That was fixed by cloning the object before mutating.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -16,6 +16,7 @@ jobs:
       # Capture the PR object and release status
       release_created: ${{ steps.release.outputs.release_created }}
       pr: ${{ steps.release.outputs.pr }}
+      prs_created: ${{ steps.release.outputs.prs_created }}
     steps:
       - name: Release Please
         id: release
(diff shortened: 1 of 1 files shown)


usefulthink:
That is incredibly helpful, thank you very much! I'll have a look to fully understand where I might have went wrong with the usePrevious implementation.

hugoatmooven:
No problem, we are in this together... I have a growing list of bugs in my backlog waiting for this fix 😂 
Let me know if there is anything I can do to help getting this published quicker.

usefulthink (approved):

usefulthink:
Your changes are making perfect sense, thanks again! This will get released with the 1.8.0 version later this week I think (1.8.0 also includes the new js-api-loader which is why I'm a bit hesitant with the release. But I'll publish a prerelease that you can already use for testing).

PR #906, opened 2026-01-28
fix: respect clickable prop in AdvancedMarker

### LINKED ISSUE #904: [Bug] AdvancedMarker remains clickable even when `clickable` is set to `false`
### Summary
In the `AdvancedMarker` component, the `clickable` option does not function as expected. Even when `clickable={false}` is explicitly passed, the marker remains clickable and intercepts the map's `onClick` events.

### Context
The current implementation in `react-google-maps/src/components/advanced-marker.tsx` is as follows:

```typescript
  useEffect(() => {
    if (!marker) return;

    const gmpClickable =
      clickable !== undefined ||
      Boolean(onClick) ||
      Boolean(onMouseEnter) ||
      Boolean(onMouseLeave);

    // gmpClickable is only available in beta version of the
    // maps api (as of 2024-10-10)
    marker.gmpClickable = gmpClickable;

    // enable pointer events for the markers with custom content
    if (gmpClickable && marker?.content && isElementNode(marker.content)) {
      marker.content.style.pointerEvents = 'all';

      if (onClick) {
        marker.content.style.cursor = 'pointer';
      }
    }
  }, [marker, clickable, onClick, onMouseEnter, onMouseLeave]);
```

### The Problem
In the logic above, gmpClickable evaluates to true if clickable is explicitly set to false, because clickable !== undefined returns true.

```TypeScript
const gmpClickable =
      clickable !== undefined || // This evaluates to true when clickable is false
      Boolean(onClick) ||
      ...
```
This prevents users from disabling marker interactions when they want the map underneath to handle the click.

### Proposed Fix
The logic should be updated to properly respect the boolean value of the clickable prop:

```TypeScript
const gmpClickable =
      (clickable !== undefined && clickable) ||
      Boolean(onClick) ||
      Boolean(onMouseEnter) ||
      Boolean(onMouseLeave);
```
I am relatively new to contributing to OSS, but I would like to help improve this library. If this proposal seems valid and aligned with the project's goals, I am happy to submit a Pull Request.

Thank you!

### Steps to Reproduce

1. Create an `AdvancedMarker` component and set the `clickable` prop to `false`.
2. Open the page in Chrome and inspect the element using Developer Tools.
3. Observe that the `gmp-clickable` attribute is still being applied to the `gmp-advanced-marker` element.

### Environment

- Library version: @vis.gl/react-google-maps@1.7.1
- Google maps version: weekly
- Browser and Version: 144.0.7559.97
- OS: Windows 11 Pro

### Logs

```text

```

shimizu-futa:
## Description
This PR fixes an issue where the `AdvancedMarker` remains clickable even when the `clickable` prop is explicitly set to `false`. 

As discussed in #904 , the current logic for `gmpClickable` incorrectly evaluates to `true` if `clickable` is defined as `false`. I have updated the logic to properly respect the boolean value of the prop.

## Related Issue
Closes #904 

## Changes
- Updated the `gmpClickable` calculation in `advanced-marker.tsx` to ensure it only evaluates to `true` when `clickable` is actually `true` (or when interaction callbacks are present).

## How to Test
1. Render an `AdvancedMarker` with `clickable={false}`.
2. Verify that the `gmp-clickable` attribute is not applied (or set to false) in the DOM.
3. Confirm that the marker no longer intercepts click events, allowing them to propagate to the map.

### DIFF
--- a/src/components/advanced-marker.tsx
+++ b/src/components/advanced-marker.tsx
@@ -301,11 +301,12 @@ function useAdvancedMarker(props: AdvancedMarkerProps) {
   useEffect(() => {
     if (!marker) return;
 
+    // when clickable is defined, we will always use its value for gmpClickable.
+    // otherwise we auto-detect based on existing event-handlers.
     const gmpClickable =
-      clickable !== undefined ||
-      Boolean(onClick) ||
-      Boolean(onMouseEnter) ||
-      Boolean(onMouseLeave);
+      clickable !== undefined
+        ? clickable
+        : Boolean(onClick) || Boolean(onMouseEnter) || Boolean(onMouseLeave);
 
     // gmpClickable is only available in beta version of the
     // maps api (as of 2024-10-10)


usefulthink (approved):
Awesome, thanks a lot.

usefulthink (approved):

PR #960, opened 2026-04-07
fix(hooks): avoid static useInsertionEffect imports

### LINKED ISSUE #957: [Bug] Build failure when importing useInsertionEffect in React <18 despite fallback to useLayoutEffect
### Description

When attempting to use useInsertionEffect with a fallback to useLayoutEffect for backward compatibility with React 16/17, the application fails at build time in environments where useInsertionEffect is not exported (React <18).

Expected behavior:
The fallback (useInsertionEffect ?? useLayoutEffect) should allow the code to run safely across React versions.

Actual behavior:
The application fails during module resolution with an error similar to:

Attempted import error: 'useInsertionEffect' is not exported from 'react'

This happens even though a runtime fallback is provided.

### Steps to Reproduce

Use a project configured with React 16 or React 17.
Add the following code:
```js
import { useInsertionEffect, useLayoutEffect } from 'react';

const useBeforeEffect = useInsertionEffect ?? useLayoutEffect;
```

Build or start the application.
Observe that the build fails with a missing export error.

### Environment

- Library version: [v1.8.1](https://github.com/visgl/react-google-maps/releases/tag/v1.8.1)
- Google maps version: [v1.8.1](https://github.com/visgl/react-google-maps/releases/tag/v1.8.1)
- Browser and Version: (any, not relevant — failure occurs at build time)
- OS: Mac OS

### Logs

This issue arises due to static analysis of ES module imports, where named imports must exist at build time. A safer pattern that works across versions is:

```js
import * as React from 'react';

const useBeforeEffect =
  React.useInsertionEffect ?? React.useLayoutEffect;
```
This avoids build-time failures by deferring resolution to runtime.

officialasishkumar:
## Summary
- stop statically importing `useInsertionEffect` so React 16/17 builds can resolve the module cleanly
- keep the React 18+ fast path by reading `useInsertionEffect` dynamically and falling back to `useLayoutEffect`
- add a bundler regression that verifies the hook still bundles when `react` does not export `useInsertionEffect`

## Testing
- `npx jest src/hooks/__tests__/use-effect-event.test.ts --runInBand`
- `npm run test:tsc`
- `npx eslint src/hooks/use-effect-event.ts src/hooks/__tests__/use-effect-event.test.ts`

Closes #957.


src/hooks/use-effect-event.ts
```diff
@@ -3,12 +3,18 @@
...
 // https://github.com/sanity-io/use-effect-event
 
-import {useInsertionEffect, useLayoutEffect, useRef} from 'react';
+import * as React from 'react';
+
+const {useLayoutEffect, useRef} = React;
+const maybeUseInsertionEffect = Reflect.get(
+  React as typeof React & Record<string, unknown>,
+  'useInsertionEffect'
+) as typeof useLayoutEffect | undefined;
```
usefulthink: this seems overly complicated, why are all those typecasts needed? Wouldn't this do the trick as well?

```tsx
const useBeforeEffect = React.useInsertionEffect ?? useLayoutEffect;
```

usefulthink (changes requested):
Thanks a lot for your contribution, I can see a lot of thought went into this.

I think the test is a bit over the top for this, it essentially just makes sure that this exact change won't be reverted, and introduces a lot of unwanted dependencies (the test-suite shouldn't need the typescript compiler and rollup to run, even if they are coincidentally installed).

src/hooks/use-effect-event.ts
officialasishkumar: Simplified to `React.useInsertionEffect ?? useLayoutEffect` and removed the bundler test. Thanks for the feedback.

usefulthink (approved):
Awesome, thanks!

src/hooks/use-effect-event.ts
```diff
@@ -3,12 +3,14 @@
...
-import {useInsertionEffect, useLayoutEffect, useRef} from 'react';
+import * as React from 'react';
+
+const {useLayoutEffect, useRef} = React;
 
 // useInsertionEffect was added in React 18; fall back to useLayoutEffect for
 // React 16/17. Both run before useEffect, so ref.current is always up-to-date
 // by the time any passive effect (or real event) reads it.
-const useBeforeEffect = useInsertionEffect ?? useLayoutEffect;
+const useBeforeEffect = React.useInsertionEffect ?? useLayoutEffect;
```
devarshiqm, line 13:
```suggestion
const useBeforeEffect =
  'useInsertionEffect' in React
    ? React.useInsertionEffect
    : React.useLayoutEffect;
```

PR #967, opened 2026-04-17
feat(map-control): add className prop

pudek357:
## Summary

Adds an optional `className` prop to `<MapControl>` so users can style the control container element. Since `MapControl` renders its children into a detached `<div>` via `createPortal` (which is then handed to the Maps JavaScript API and positioned by the control-layout algorithm), there is currently no way for consumers to target that wrapping element from the React tree - the new prop closes that gap.

The implementation mirrors the existing `className` handling in sibling components (`AdvancedMarker`, `InfoWindow`): an effect syncs the prop to the DOM node as `controlContainer.className = className ?? ''`.

## Changes

- `src/components/map-control.tsx` — add optional `className` prop and a `useEffect` that syncs it to the portal container.
- `src/components/__tests__/map-control.test.tsx` — add unit tests covering initial application, updates, and unset/removal of `className`.
- `docs/api-reference/components/map-control.md` — document the new prop under an "Optional" section.

## Test plan

- [x] `npm run test:linter` passes
- [x] `npm run test:prettier` passes
- [x] `npm run test:tsc` passes
- [x] `npm run test:unit -- map-control` — 3/3 tests pass (1 existing + 2 new)
- [x] Full `npm run test:unit` suite: no new failures introduced (pre-existing `ts-jest` failures on `main` are unrelated to this change)

## Notes for reviewers

- No breaking changes; prop is optional and defaults to unset.
- API mirrors `className` on other components in the library for consistency.
- Happy to open a Discussion first if the maintainers prefer that flow for small additive props like this one — just let me know.

### DIFF
--- a/docs/api-reference/components/map-control.md
+++ b/docs/api-reference/components/map-control.md
@@ -45,5 +45,14 @@ const App = () => (
 The position is specified as one of the values of the `ControlPosition` enum, which
 is an exact copy of the [`google.maps.ControlPosition`][gmp-ctrl-pos] type.
 
+### Optional
+
+#### `className`: string
+
+A CSS class name applied to the container element that wraps the control content
+on the map. This is useful for styling or targeting the control from outside,
+since the children are rendered into a container that is managed by the Maps
+JavaScript API rather than directly into the React tree.
+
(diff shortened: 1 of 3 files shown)


mrMetalWood:
Thanks for this! Looking good.

PR #970, opened 2026-04-24
fix: prevent api provider race condition, defer loading state update

### LINKED ISSUE #968: [Bug] APIProvider 1.8.x crashes with "google.maps.Settings is undefined" due to race in LOADED fast-path
### Description

Since v1.8.0 (PR #913), `APIProvider` runs an effect that unconditionally calls `google.maps.Settings.getInstance()` once `status === APILoadingStatus.LOADED`:

```ts
// src/components/api-provider.tsx
useEffect(() => {
  if (status !== APILoadingStatus.LOADED) return;

  const settings = google.maps.Settings.getInstance();
  if (fetchAppCheckToken) {
    settings.fetchAppCheckToken = fetchAppCheckToken;
  } else if (settings.fetchAppCheckToken) {
    settings.fetchAppCheckToken = null;
  }
}, [status, fetchAppCheckToken]);
```

In the "API already available" fast-path in `useGoogleMapsApiLoader`, `status` is set to `LOADED` **before** `importLibrary('core' / 'maps')` has resolved:

```ts
if (window.google?.maps?.importLibrary as unknown) {
  if (!serializedApiParams) {
    updateLoadingStatus(APILoadingStatus.LOADED); // <-- fires the Settings effect
  }
  await Promise.all(
    librariesToLoad.map(name => importLibraryCallback(name))
  );
  ...
}
```

When the Settings effect runs, `google.maps.importLibrary` exists (the bootstrap installed it) but the actual `maps/api/js` script has not finished loading yet, so `google.maps.Settings` is `undefined` and `Settings.getInstance()` throws:

```
TypeError: can't access property "getInstance", google.maps.Settings is undefined
```

This reproduces reliably in production builds and is mostly invisible in dev, presumably because of scheduling slack and StrictMode double-mount timing differences.

The effect runs regardless of whether `fetchAppCheckToken` is provided (the effect has no early-return on `fetchAppCheckToken === undefined`), so there is no userland workaround — omitting the prop does not skip `Settings.getInstance()`.

### Suggested fix

Defer the `LOADED` status update until `importLibrary` actually resolves in the fast-path, e.g.:

```ts
if (window.google?.maps?.importLibrary as unknown) {
  await Promise.all(
    librariesToLoad.map(name => importLibraryCallback(name))
  );
  if (!serializedApiParams) {
    updateLoadingStatus(APILoadingStatus.LOADED);
  }
  if (onLoad) onLoad();
  return;
}
```

Or, as a defensive fallback, guard the Settings effect:

```ts
if (!google.maps.Settings) return;
```

### Steps to Reproduce

1. Mount `<APIProvider apiKey={...}>` at the root of a production Vite + React 19 app (SSR or CSR both reproduce).
2. Don't pass `fetchAppCheckToken`.
3. Observe `TypeError: can't access property "getInstance", google.maps.Settings is undefined` shortly after the provider mounts.

Also reproducible when another component has already caused `google.maps.importLibrary` to be installed (for example `@react-google-maps/api`'s `useJsApiLoader`) but the maps script has not finished loading yet.

### Environment

- Library version: @vis.gl/react-google-maps@1.8.0 – 1.8.3 (bisected to PR #913)
- Google maps version: weekly (default, no `version` prop passed)
- Browser and Version: Firefox 145, Chrome 144 (any)
- OS: Linux, macOS

### Logs

```text
TypeError: can't access property "getInstance", google.maps.Settings is undefined
    at APIProvider effect (api-provider.tsx:374)
```

### Workaround

Pin `@vis.gl/react-google-maps` to `1.7.1`, which predates PR #913.

mrMetalWood:
Reproduced and confirmed problem and solution of #968 and applied suggested fix

## Summary

Fixes an `APIProvider` race when `google.maps.importLibrary` is already installed by an external loader, but the Maps JavaScript API has not finished loading yet.

Previously, the external-loader fast path marked the API as `LOADED` before awaiting `importLibrary('core')` and `importLibrary('maps')`. This allowed provider effects and child components to run too early, causing errors such as `google.maps.Settings is undefined` or `google.maps.Map is not a constructor`.

## Changes

- Wait for required libraries to resolve before setting `APILoadingStatus.LOADED` in the external-loader fast path.
- Add a regression test for externally installed `google.maps.importLibrary` with `google.maps.Settings` initially unavailable.

closes #968

### DIFF
--- a/src/components/__tests__/api-provider.test.tsx
+++ b/src/components/__tests__/api-provider.test.tsx
@@ -244,6 +244,45 @@ test('sets fetchAppCheckToken on google.maps.Settings after API loads', async ()
   expect(settingsInstance.fetchAppCheckToken).toBe(mockFetchToken);
 });
 
+test('waits for externally installed importLibrary before marking API as loaded', async () => {
+  const mockFetchToken = jest.fn().mockResolvedValue({token: 'test-token'});
+  const getInstanceMock = jest.fn(() => settingsInstance);
+
+  google.maps.importLibrary = jest.fn(async () => {
+    await importLibraryPromise;
+
+    return {} as google.maps.MapsLibrary;
+  }) as typeof google.maps.importLibrary;
+  google.maps.Settings = undefined as unknown as typeof google.maps.Settings;
+
+  render(
+    <APIProvider apiKey={'apikey'} fetchAppCheckToken={mockFetchToken}>
+      <ContextSpyComponent />
+    </APIProvider>
+  );
+
+  await waitFor(() =>
+    expect(ContextSpyComponent.spy.mock.lastCall[0].status).toBe(
+      APILoadingStatus.NOT_LOADED
+    )
+  );
+
+  await act(async () => {
+    google.maps.Settings = {
+      getInstance: getInstanceMock
+    } as unknown as typeof google.maps.Settings;
+    triggerMapsApiLoaded();
+  });
+
+  await waitFor(() =>
+    expect(ContextSpyComponent.spy.mock.lastCall[0].status).toBe(
+      APILoadingStatus.LOADED
+    )
+  );
+  expect(getInstanceMock).toHaveBeenCalled();
+  expect(settingsInstance.fetchAppCheckToken).toBe(mockFetchToken);
+});
+
 describe('internalUsageAttributionIds', () => {
   test('provides default attribution IDs in context', () => {
     render(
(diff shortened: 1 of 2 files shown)


usefulthink (approved):
Thanks! Seems like a simple enough fix that we can release it right away.

cramt:
amazing, thanks for the quick fix :heart:

PR #973, opened 2026-04-27
chore/update gmp types

usefulthink:
- **chore: update dependencies including the @types/google.maps package**
- **chore: disaqble new linting rules after eslint-update**


rodrigosantos-gympass (approved):

usefulthink (commented):
Looks good to me. Except the example tsconfig importing from outside the example folder, but we can fix that later.

PR #989, opened 2026-05-21
feat: add map-3d-route example with 3D Map and custom route element (v=alpha)

anglarett:
gmp-route-3d
routes-3d-polyline


mrMetalWood:
Hi @anglarett 

Thanks for the PR. Looks good to me :) 

A few things i noticed:
- I had trouble running the examples locally with the `package.json` and `vite.config.js` setup. I think we should stick to the setup, just like the other examples. No need for additional alias in the vite.config.js. And there was an old vite (v5) version in the package.json.
- We would need two screenshots or preview images of the examples for the example gallery

I could do the finishing touches on this if you want, otherwise it would be great if you go over it for a final pass.

Issue #720, opened 2025-03-22
Issue with using route API


EdwardGrasshopper:
### Description

I got this error when I tried making a route from an origin to a destination.
**MapsRequestError: DIRECTIONS_ROUTE: REQUEST_DENIED: There was an issue performing a Directions request.**

I just started to use my Google Maps API and did notice the issue located at this (https://issuetracker.google.com/issues/255359343), and I currently have my API key set to unrestricted, but I'm curious if this behavior is expected while the issue is pending?

I also got these errors too:
**Uncaught (in promise) MapsRequestError: DIRECTIONS_ROUTE: REQUEST_DENIED: The webpage is not allowed to use the directions service.**

and 

**Directions Service: This API project is not authorized to use this API.  For more information on authentication and Google Maps JavaScript API services please see: https://developers.google.com/maps/documentation/javascript/get-api-key**

I can provide more info if necessary.

### Steps to Reproduce

Use the useMap hook and the useMapsLibrary hook using 
const map = useMap();
const routesLibrary = useMapsLibrary("routes");

I followed all the general steps used in this video put out by Google Maps:
https://www.youtube.com/watch?v=tFjOIZGCvuQ&list=PL2rFahu9sLJ2QuJaKKYDaJp0YqjFCDCtN&index=3

### Environment

- Library version:
- Google maps version: weekly
- Browser and Version:
- OS:

### Logs

```text

```

usefulthink:
What is currently happening is that the Google Maps team is in the process of moving the Directions- and DistanceMatrix APIs into the 'legacy' state. This means that usage in new projects is discouraged (officially no longer possible, but there seems to be a way, see below), but they continue to be supported as-is for existing projects: https://developers.google.com/maps/legacy

The main takeaway here is this: While there already is the library named "routes", it doesn't use the new Routes API, but rather the Directions API, which is now in legacy. There currently is no client library for using the routes API.

So if you want to use the Routes API instead of the Directions API ([and there are good reasons to](https://developers.google.com/maps/documentation/routes/migrate-routes-why)), you can either call the routes API on the server and use the `@googlemaps/routing` package, or you have to call it via fetch on the client.

If not, you should still be able to enable the legacy directions API by using this direct link: https://console.cloud.google.com/apis/library/directions-backend.googleapis.com (however, that might just be a transitional thing that could stop working at any time).

raffidahmad:
https://developers.google.com/maps/documentation/routes/migrate-routes

[...]

PR #991, opened 2026-05-22
examples(directions): upgrade Directions example to modern Routes API.

caio1985:
### Description
This PR resolves issue #720 by upgrading the official `directions` example component to utilize the modern, non-deprecated client-side **Routes API (`google.maps.routes.Route`)** instead of the deprecated legacy `DirectionsService` and `DirectionsRenderer`.

### Key Improvements
1. **Bypasses CORS Limits Natively:** Sourced client-side routing via the JavaScript SDK library engine (`Route.computeRoutes`), completely avoiding browser CORS restrictions that block direct REST API calls.
2. **Vibrant Modern Styling:** Custom-styled route polylines using the modern `Route.createPolylines()` with clean React-friendly rendering options.
3. **Eliminates Deprecation Warnings:** Cleaned up console health by eliminating legacy service runtime warnings.
4. **Refactored Control Panel:** Updated description texts, headers, and links to reflect the recommended Routes API upgrade while retaining clear migration instructions for legacy service users.

### Verification & Testing
- Run typecheck compiler: `npm run test:tsc:examples` (compiles with **0 errors and 0 warnings**).
- Serves locally: `npm run start-local` inside `examples/directions` works perfectly and draws the route beautifully.

image


mrMetalWood:
Hi @caio1985 👋 

Thanks for the PR! Overdue that we migrate the legacy directions example.

A few things to finalize this PR:
- The official Google Maps typescript types recently got an update. Lets try to use those types as much as possible. They should also include typings for the Routes API. That way we dont have to create our own and can avoid any "anys" as much as possible
- Would be nice to also have the origin and destination marker still visible like previously. You can get those with `<polyline>.createWaypointAdvancedMarkers()` method.
- The routes-panel is not displayed anywhere as far as i can see  👀 Either show it somewhere or just return `null` from there. I dont think we necessarily need the distance/time values displayed anywhere.

Would you be able to go for another pass on this? If not, I could also take it over from here.

PR #1023, opened 2026-07-08
Guard against invalid camera values in map events

tzh476:
## Summary

This prevents invalid camera values returned by the Maps API from being turned into fallback camera state inside the Map component.

In particular, when a `bounds_changed` event reads an invalid camera snapshot such as a non-finite zoom value, the component now:

- skips updating the internal tracked camera state instead of overwriting it with fallback values like `zoom: 0`
- skips emitting an `onCameraChanged` event with synthetic fallback bounds/center/zoom
- keeps the existing warning so unexpected Maps API values remain visible to developers

This is a defensive fix for one feedback-loop failure mode discussed in #563. It does not claim to fix every underlying Maps JS API pinch-zoom issue, but it avoids amplifying invalid camera snapshots through React state and controlled camera updates.

## Tests

- `npm test`

### DIFF
--- a/src/components/__tests__/map.test.tsx
+++ b/src/components/__tests__/map.test.tsx
@@ -350,6 +350,39 @@ describe('camera configuration', () => {
 
     expect(mapInstance.moveCamera).not.toHaveBeenCalled();
   });
+
+  test('tracked camera state accepts valid camera values while bounds are unavailable', async () => {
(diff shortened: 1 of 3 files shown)


usefulthink:
Thanks so much! I've been wanting to figure out a better solution for a while now..

PR #1052, opened 2026-08-10
feat(map-control): add style prop

### LINKED ISSUE #379: Add className and style prop to the MapControl component
### Target Use Case

Currently, the MapControl component does not support applying custom styles or classes, which limits its flexibility. We need the ability to add className and style props to this component to allow for custom styling. For example, we want to make the control take 100% of the container by using `inset: 0px 0px auto 0px;`

### Proposal

Modify the MapControl component to accept className and style props.

samithahansaka:
Adds a `style` prop to `MapControl`, applied to the same container element that already receives `className`.

Closes #379. The `className` half of that issue shipped in #967, so this is the remaining part.

### Motivation

The original request was to let a control break out of the default layout, for example `{inset: '0px 0px auto 0px'}` to make it span the full width of the map. Without this, the container the control renders into cannot be styled at all from React, because it is created imperatively and handed to the Maps JavaScript API.

### Implementation

It follows the pattern already used by `InfoWindow` and `Popover`:

- `setValueForStyles` from `src/libraries/set-value-for-styles.ts`, with a `prevStyleRef` holding the previously applied styles
- inside a `useLayoutEffect`, so the control is not painted unstyled for a frame
- `className` moved into the same effect, matching `info-window.tsx`

Using the shared helper means the prop behaves exactly like React's `style` everywhere else in the library: numbers get an implicit `px` suffix, CSS custom properties work, and falsy values unset a property. It also only touches the properties actually passed in, so inline styles the Maps API writes on the container are left alone.

### Tests

Five new tests covering application, updates, removal of properties dropped between renders, `px` suffixing, custom properties, and that unrelated inline styles survive an update.

`npm test` and `npm run build` pass.

### Two things worth discussing

**Interaction with control layout.** You raised this in the issue: controls registered at the same `position` are laid out by the API, so styles that change size or placement will affect neighbouring controls. I documented that caveat rather than trying to prevent it, since preventing it would mean deciding which properties are allowed. Happy to restrict the prop if you would rather.

**Two eslint-disable comments.** `react-hooks/immutability` reports the container mutation twice, once for the indirect write through `setValueForStyles` and once for the direct `className` assignment, so both needed suppressing. If there is a preferred way to structure this so the rule is satisfied without the second comment, I am glad to change it.

### DIFF
--- a/docs/api-reference/components/map-control.md
+++ b/docs/api-reference/components/map-control.md
@@ -54,5 +54,16 @@ on the map. This is useful for styling or targeting the control from outside,
 since the children are rendered into a container that is managed by the Maps
 JavaScript API rather than directly into the React tree.
 
+#### `style`: CSSProperties
+
+Inline styles applied to the same container element. Useful when the control has
+to break out of the default layout, for example `{inset: '0px 0px auto 0px'}` to
+make it span the full width of the map.
+
+Note that the container is positioned by the Maps JavaScript API, which lays out
+all controls registered for the same `position`. Styles that change the size or
+placement of the container can therefore affect how neighbouring controls are
+arranged.
+
 [gmp-custom-ctrl]: https://developers.google.com/maps/documentation/javascript/controls#CustomControls
 [gmp-ctrl-pos]: https://developers.google.com/maps/documentation/javascript/controls#ControlPositioning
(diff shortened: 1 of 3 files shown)


usefulthink (approved):
looks good. Thanks a lot!

PR #1054, opened 2026-08-11
chore: remove unused clustering hooks from src

samithahansaka:
Removes `src/hooks/use-supercluster-worker.ts` and `src/hooks/use-map-viewport.ts`, as discussed in #1049.

Both were added alongside the worker clustering example in #891. The example imports its own copies under `examples/worker-marker-clustering/src/hooks`, so the ones in `src` were never used by anything.

### Why this is not a breaking change

- neither is re-exported from `src/index.ts`
- neither is imported anywhere inside `src`
- no tests and no docs reference them
- `supercluster` is not a dependency of the library
- they never reached `dist`, so no bundle ever contained them

`./src` is part of `files`, so the sources did ship inside the npm tarball. They still were not reachable: the `exports` map only declares `.`, `./server`, `./3d` and `./examples.css`, so there is no specifier that resolves to them.

### Nothing is lost

Both deleted files are byte-identical to the copies under `examples/worker-marker-clustering/src/hooks`, apart from one import path (`./use-map` versus `@vis.gl/react-google-maps`) and a doc link. I diffed them to confirm.

Worth noting that `useMapViewport` existed in three places, not two. `examples/custom-marker-clustering` has its own older copy which this PR does not touch, so that one is now the remaining duplicate.

`npm test` and `npm run build` pass, examples included.


usefulthink (approved):
Awesome, thanks a lot!

Issue #524, opened 2024-09-10
[Feat] improve performance significantly by selectively updating the Polygon options


nimlevy:
### Target Use Case

We are using the library to give our clients the ability to select zip code areas on the Google Map. Imagine a map marked with zip code boundaries, the user can hover to see details and then click to select the zip code (changing opacity and adding to a list)

### Proposal

The issue (somewhere between a bug and a feature) is that the dependencies of the [Polygon example](https://github.com/visgl/react-google-maps/blob/main/examples/geometry/src/components/polygon.tsx) (and probably the other geometry components as well) where the `polygon.setOptions` is called, are setting the options every time the mouse hovers over the polygon's area.
Updating the dependencies to the specific properties that would affect the rendering of the polygon (e.g. the `strokeOpacity`, `fillOpacity`, etc.) drastically improves performance.

I am not sure what is the best way to go about it, but the assumption stated in the following comment is, unfortunately, incorrect.
```
// update PolygonOptions (note the dependencies aren't properly checked
// here, we just assume that setOptions is smart enough to not waste a
// lot of time updating values that didn't change)
```
In other words - the `setOptions` is **not** smart enough.

Note: I have seen a drastic decrease in performance (color changes, info window's movement, etc. were lagging badly) when using the original format of paths (array of lat & lng) that React's dependency check could not properly compare

usefulthink:
Like I said in #379, any help is very much appreciated! 

Especially since we will eventually move the geometry components (Polygon, Polyline and Circle) into the main library.

nimlevy:
@usefulthink I'm not sure why, but I am not seeing the comment made by JalilSeif
But if they have any questions, or can use my code snippets, I can help.

usefulthink:
Yeah, that must've been deleted in the meantime. It's a bit confusing that GitHub doesn't show that there was a deleted comment...

samithahansaka:
I have this working locally, but there is a contract question I would rather settle with you before opening a PR.

First, a correction to the issue as written. The part about `setOptions` running on every hover is already fixed: the options object is now wrapped in `useMemoized(..., isDeepEqual)`, so an unchanged render never reaches `setOptions`. What is still there is that when one value does change, the whole options object is sent. Changing `fillOpacity` on hover also re-sends `editable`, `draggable`, `clickable` and every stroke and fill value, and writing `editable` makes the API rebuild the vertex and midpoint handles. That is the part that is expensive on shapes with many vertices.

[...]

PR #1061, opened 2026-08-21
perf(geometry): only send changed options to the maps api

samithahansaka:
Makes `Polygon`, `Polyline`, `Circle` and `Rectangle` send only the options that actually changed to `setOptions`.

Refs #524, and follows the discussion there.

### What was already fixed

The issue reported `setOptions` running on every render. That is no longer true: the options object is wrapped in `useMemoized(..., isDeepEqual)`, so an unchanged render never reaches `setOptions`.

### What is still happening

When one value changes, the whole options object is still sent. Changing `fillOpacity` on hover also re-sends `editable`, `draggable`, `clickable` and every stroke and fill value. Writing `editable` or `draggable` makes the API rebuild the vertex and midpoint handles, which is the cost the original report described on shapes with many vertices.

### The change

`useAppliedOptions` keeps the previously applied options per instance and passes only the changed subset. This is equivalent, because the API leaves options that are not specified untouched.

It replaces four copies of the same effect, so the diff removes duplication as well as adding the behaviour. The pure diffing part lives in `src/libraries/get-changed-options.ts` next to the other helpers, and both symbols are marked `@internal`.

It also applies options **once** on mount instead of twice. The instances are already constructed with the full option set, so the creation effect records what it applied and the hook starts from that, rather than immediately re-sending everything. For an externally provided `polygon` or `polyline` that was two full applies per mount, three under StrictMode.

### On the contract change

This gives up something the current code does incidentally: today any options change re-asserts every value, so an instance mutated from outside gets snapped back on the next update. With a diff it does not.

I raised that in #524 and @usefulthink's answer was that the internal design does not need to accommodate code mutating objects it does not own. This PR follows that, diffing everything rather than special-casing which keys are expensive. Flagging it here so the decision is visible in the PR too, not only in the issue.

### Tests

Twelve for the hook and helper, plus one per component and two mount tests. I checked that each new component test fails against `main` and passes with the change, so they test the new behaviour rather than restating the old.

`npm test` and `npm run build` pass: 190 tests, 22 suites.

### What I could not verify

**No benchmark.** Measuring this properly needs a real browser, an API key and a large polygon. The numbers in the original report predate the `useMemoized` fix, so they no longer describe the current code. What holds without measuring is that the change strictly reduces the work handed to the API and cannot be slower. The `editable` handle rebuild is the mechanism I am reasoning from, not something I profiled.

### Related, not in this PR

`marker.tsx` and `use-map-options.ts` still carry the "we just assume that `setOptions` is smart enough to not waste a lot of time updating values that didn't change" comment. The hook is generic and would apply to both. `useMapOptions` would need more care, since `map/index.tsx` writes to the same instance separately. Happy to follow up if wanted.


src/components/polygon.tsx
```diff
@@ -102,6 +104,11 @@ function usePolygon(props: PolygonProps) {
     isDeepEqual
   );
 
+  // options the instance was constructed with, so they are not applied twice
+  const appliedOnCreateRef = useRef<AppliedOptions<
```
usefulthink: I don't like that this additional ref is needed in all geometry components for the hook to work. A better solution would be to handle the initial state within the hook. The hook can track for itself if it was the first time being called, and treat the passed options in that case as the initial value without applying them.

src/hooks/use-applied-options.ts
```diff
@@ -0,0 +1,61 @@
...
+ * This tracks what it has written rather than reading the instance back, so it
+ * assumes nothing else writes these options on the same instance. A value set
+ * from outside is not restored on the next update.
+ *
+ * `appliedOnCreateRef` lets the caller declare the options an instance was
+ * already constructed with, so they are not written a second time on mount.
+ *
+ * @internal
+ */
+export function useAppliedOptions<TOptions extends object>(
```
usefulthink: I would change the name to better capture what it does, like `useApplyChangedOptions` or `useOptionsUpdater`.

src/hooks/use-applied-options.ts
```diff
@@ -0,0 +1,61 @@
...
+    // instance, either by an earlier run or by whoever created it
+    let prevOptions: TOptions | null = null;
+
+    if (prevRef.current?.instance === instance) {
+      prevOptions = prevRef.current.options;
+    } else if (appliedOnCreateRef?.current?.instance === instance) {
+      prevOptions = appliedOnCreateRef.current.options;
+    }
+
+    const changedOptions = getChangedOptions(options, prevOptions);
```
usefulthink: To correctly track options being removed, we should accumulate the latest state for all options that have been written to the object by this hook and not just the last state.

usefulthink (changes requested):
Thank you very much for your work! I'd love to work with you on this, but if you don't have the time at the moment, let me know and I'll take it from here.

The PR is definitely going in the right direction and adds a behavior we want in the library: geometry components shouldn't resend the full options object to the Maps API with every change.

A few suggestions to make the implementation clearer and more robust:

1. **Rename the hook**
   
   I’d suggest renaming `useAppliedOptions` to something like `useOptionsUpdater`.
   
   That name is immediately understandable, fits the mandatory hook `use` prefix naturally, and describes the role of the hook without exposing too much implementation detail.

2. **Move tracking fully into the hook**

   The current `appliedOnCreateRef` pattern spreads the tracking logic across the component setup code and the hook. I’d prefer the hook to own this completely.

   The component invariant can be:

   > Before an instance is stored in state / exposed to `useOptionsUpdater`, the initial options have already been applied.

   This is already true for internally created geometry instances because options are passed to the constructor. For externally supplied instances, we should explicitly apply the initial options once before storing the instance.

   With that invariant, the hook can handle this internally:

   - if it's is the first time the instance is seen, store the current options as tracked state and don't call `setOptions`;
   - on later updates, compute and apply only the changed subset.

3. **Use a `WeakMap` for per-instance tracking**
   
   We can safely store the tracked options per object in a module-local `WeakMap` that holds a copy of the options as we expect them to look:
   
   ```typescript
   WeakMap<Instance, trackedOptions>
   ```
   
   This avoids strong references to Maps instances and keeps cleanup memory-safe without having to keep track of the objects.

4. **Track accumulated applied options, not only the previous render’s options**
   
   The tracked state should represent the options the hook believes are currently applied to the instance, which enables better handling of undefined values.
   If an option was previously applied and then disappears from the next options object, it should be explicitly unset.
   Similarly, if an option appears with an explicit `undefined` value, that should still be applied once. This avoids stale options accumulating on the Maps instance and better matches React prop semantics.
    
_**TL;DR:**_ Overall, I’d keep the core idea, but simplify the API around it: `useOptionsUpdater(instance, options)`, internally backed by a `WeakMap`, with first-seen instance tracking and explicit unsetting of removed options.

This would be a suggestion for the general structure of the `useOptionsUpdater` hook based on this:

```tsx
import {useEffect} from 'react';
import {deepEqual as isDeepEqual} from 'fast-equals';

type OptionsInstance<TOptions extends object> = {
  setOptions(options: Partial<TOptions>): void;
};

const trackedOptionsByInstance = new WeakMap<object, object>();

export function useOptionsUpdater<TOptions extends object>(
  instance: OptionsInstance<TOptions> | null,
  options: TOptions
): void {
  useEffect(() => {
    if (!instance) return;

    const trackedOptions = trackedOptionsByInstance.get(instance) as
      | Partial<TOptions>
      | undefined;

    // First time we see this instance:
    // assume the current options were already applied by construction/setup.
    if (!trackedOptions) {
      trackedOptionsByInstance.set(instance, {...options});
      return;
    }

    const changedOptions = getChangedOptions(options, trackedOptions);

    if (!changedOptions) return;

    instance.setOptions(changedOptions);

    trackedOptionsByInstance.set(
      instance,
      applyOptionsUpdate(trackedOptions, options)
    );
  }, [instance, options]);
}

function getChangedOptions<TOptions extends object>(
  nextOptions: TOptions,
  trackedOptions: Partial<TOptions>
): Partial<TOptions> | null {
  let changedOptions: Partial<TOptions> | null = null;

  const keys = new Set<keyof TOptions>([
    ...(Object.keys(trackedOptions) as Array<keyof TOptions>),
    ...(Object.keys(nextOptions) as Array<keyof TOptions>)
  ]);

  for (const key of keys) {
    const hasNextValue = key in nextOptions;

    if (!hasNextValue) {
      changedOptions ??= {};
      changedOptions[key] = undefined;
      continue;
    }

    if (
      key in trackedOptions &&
      isDeepEqual(trackedOptions[key], nextOptions[key])
    ) {
      continue;
    }

    changedOptions ??= {};
    changedOptions[key] = nextOptions[key];
  }

  return changedOptions;
}

function applyOptionsUpdate<TOptions extends object>(
  trackedOptions: Partial<TOptions>,
  nextOptions: TOptions
): Partial<TOptions> {
  const updatedOptions: Partial<TOptions> = {...trackedOptions};

  for (const key of Object.keys(trackedOptions) as Array<keyof TOptions>) {
    if (!(key in nextOptions)) {
      delete updatedOptions[key];
    }
  }

  for (const key of Object.keys(nextOptions) as Array<keyof TOptions>) {
    updatedOptions[key] = nextOptions[key];
  }

  return updatedOptions;
}
```

samithahansaka:
Thanks, this is a better shape than what I had. I am on it, no need to take it over. All four points are done, in separate commits so the diff since your review stays readable.

One thing you should know: the first-seen idea does not work, and I want to show you why rather than hide it in the diff.

### The problem

"The options are already applied before the instance reaches the hook" is true inside the creation effect. It is not true across renders.

Parent effects run after child effects. So if a parent sets state in its own mount effect, that update lands in the same commit that stores the instance:

```tsx
const Parent = () => {
  const [color, setColor] = useState('#ff0000');
  useEffect(() => setColor('#00ff00'), []);
  return <Circle center={c} radius={100} fillColor={color} />;
};
```

The constructor gets red. The hook then sees green, assumes green was already applied, and writes nothing. The circle stays red for good.

On `main` this correctly calls `setOptions({fillColor: '#00ff00', ...})`.

It is worse with `editable`. If that is the option arriving late, there are no `setOptions` calls at all, so the vertex handles are never built and `onPathsChanged` stops working with no error.

This is not an unusual setup. An async map, or any parent effect, gets you there.

### The fix

Record the options where they are actually applied, instead of guessing.

There is now a `markOptionsApplied(instance, options)` next to the hook, called right after the constructor or the external `setOptions`. Everything you asked for stays: the `WeakMap`, the tracking inside the hook, and no extra parameter on the hook itself.

If an instance has no recorded options, it now gets the full set. So forgetting to record is safe: you get a redundant write, not a missing one.

The example above is now a test.

### Three smaller fixes

- `applyOptionsUpdate` did nothing. It worked out to `{...nextOptions}`, because the second loop overwrote what the first deleted. Removed.
- `key in` reads inherited properties, so a `toString` key slipped through into `setOptions`. Now `Object.hasOwn`, same as `set-value-for-styles.ts`.
- The tracked copy was shallow, so an `icons` array mutated in place was compared against itself. If any other option changed in the same render, the icon change was dropped and only the other option went through. It is a deeper copy now.

### Two questions for you

**1. Should unsetting write `null` or `undefined`?**

Your sketch uses `undefined`. I used `null`, because the option types declare every property as `T | null`, and because `toHaveBeenCalledWith` ignores undefined values, which makes the behaviour hard to test.

But `null` might be wrong. `visible` defaults to `true`, and `fillOpacity: null` reads as 0. So removing a prop could hide the shape instead of restoring the default.

I cannot test this without a real API key. Do you know what the API does? It is a one-line change either way.

**2. What if two components share one external instance?**

The `WeakMap` is keyed by the instance, so two components pointing at the same one overwrite each other's baseline. Rendering `<Polygon polygon={ext} strokeColor="#ff0000" />` next to `<Polygon polygon={ext} />` gives four writes:

```
{strokeColor: '#ff0000', clickable: false, draggable: false, editable: false}  // A creates
{clickable: false, draggable: false, editable: false}                          // B creates
{strokeColor: '#ff0000'}                                                       // A re-applies
{strokeColor: null}                                                            // B unsets it
```

So they fight, and the last writer wins. Same if an instance moves between elements in a list reorder.

On `main` both components just write their full set, so the shape ends up with whatever the last one declared and stays there. Is this worth handling, or out of scope for the `polygon` and `polyline` props?

195 tests passing, and the component tests still fail against `main`, so they are testing the change rather than restating what was already there.

usefulthink:
Excellent observation – that flaw slipped past me. I'm still wondering if there's a way to do this without needing the second function. For some reason it doesn't feel right that we need a separate function to do this correctly. 

> 1. Should unsetting write null or undefined?

I think null is perfectly fine and should also indicate "use the default" to the maps API; it's probably even better.

> 2. What if two components share one external instance?

The question is, can we think of a case where a user might think this is what they need? 
The only way to get there would be if the user passes the same API object to multiple instances of the Polygon/Polyline component. I think we can ignore that for now.

usefulthink:
Another thought: maybe the best approach is to not use a hook at all and instead use a wrapper/helper function (e.g. `setMapObjectOptions(object, options)` on every render / `setMapObjectOptions(object, options, {alreadyApplied: true})` for new objects) that handles filtering already applied options. That'd then be a 1:1 replacement for `object.setOptions(options)`. Architecturally, that could be translated to all sorts of other API surfaces if there's a friction caused by the React way of doing things.

samithahansaka:
Done, that shape is better. There is now one exported function and no hook:

```ts
setMapObjectOptions(object, options)
setMapObjectOptions(object, options, {alreadyApplied: true})
```

It lives in `src/libraries/` rather than `src/hooks/` since it is no longer React-aware, and it is generic over `{setOptions(options)}`, so it would drop into `Marker` or `useMapOptions` unchanged if that turns out to be worth doing.

Your other two answers needed no code. It already writes `null`, and I have left the shared-instance case alone.

One tradeoff to flag, since it cuts against your first review comment. Without the hook, each component now has its own four-line effect, so there are four near-identical effects where there was one shared call. I did it your way rather than keeping a hook wrapper on top, but say the word if you would rather have the wrapper back.

### Two bugs I introduced and have now fixed

Worth calling out rather than leaving in the diff.

**`Object.hasOwn` is ES2022.** I used it and told you it matched `set-value-for-styles.ts` next door. That file uses `hasOwnProperty`, so I cited the precedent and then did not follow it. `tsc` downlevels syntax but not built-ins, so it was being emitted verbatim into `dist` against a target of `ES2015` and would have thrown on Safari below 15.4 and Node below 16.9 the first time a shape updated. Now uses the same `hasOwnProperty` call as its neighbour.

**Two `as never` casts** in the polyline test, on the one component-level test that covers the deep copy. `IconSequence` has no required fields, so the prop type accepts the value as written and the cast bought nothing except making it assignable to anything. The test could have kept passing while checking nothing.

Also hardened two edge cases in the diff: the accumulator inherited from `Object.prototype`, so an options bag with an own `__proto__` key hit the inherited setter and produced a truthy result with no own keys, writing `{}` on every render; and the snapshot was copying null-prototype values into a plain object, which `fast-equals` then reported as changed every time.

### One question left

An explicit `undefined` and a dropped prop now produce different values. `fillColor={cond ? '#f00' : undefined}` keeps the key, so it forwards `undefined`. `{...(cond && {fillColor: '#f00'})}` drops the key, so it writes `null`. In JSX those read as the same intent, and prop spreading turns one into the other.

Since you have confirmed `null` means "use the default", I think an explicit `undefined` should also write `null` rather than being forwarded. That is a one-line change, but it makes `undefined` and absent behave identically, so I would rather you confirm than assume.

194 tests. The component tests still fail against `main` and pass here.

usefulthink (approved):
I think it's fine as it is. I don't actually *know* if `undefined` and `null` behave exactly the same in the maps API, but I assume they do so this should be fine.

usefulthink:
Again, thank you very much for your time and work on this, and welcome as a new contributor :D

PR #1074, opened 2026-09-15
fix(hooks): skip writing props the consumer never provided

rpinckne:
## Summary

`usePropBinding` assigns every bound prop as soon as the element exists, including the props the consumer left out, so each custom element receives a burst of `undefined` writes on mount. Two effects of that on the Maps API:

- it clobbers defaults the element set up for itself (the class of bug #867 / #1066 fixed at one call site);
- for deprecated or channel-gated properties it logs a warning the consumer never caused. On the quarterly channel every `<Polyline3D>` logs `autofitsCamera is not supported in this version` and `The coordinates property is deprecated`, and every `<Polygon3D>` adds `innerCoordinates` and `outerCoordinates`, once per element, even when the app only ever passed `path` and `innerPaths`.

## Change

A prop is written only once it has a value. After that, a later `undefined` still clears it (removing a prop keeps working), and a fresh object starts untouched even if the hook wrote to a previous one.

This is the general form of the guard #1066 adds inside `useAdvancedMarkerAnchoring`; it covers every `usePropBinding` call site in the 3D components, `Marker3D`, and `Popover`.

## Tests

- `src/hooks/__tests__/use-prop-binding.test.tsx`: unprovided prop never written (no own property), provided values including `null` written, value-then-`undefined` clears, and a new object is left untouched.
- `npm test`: 23 suites / 194 tests pass, `tsc --project tsconfig.test.json`, eslint and prettier clean.


usefulthink:
Awesome work, thanks for spotting and fixing this!

